PR #4, opened 2026-02-07
feat(pkg/secrets): add AES-256-GCM encryption package

dmitrymomot:
## Summary

This PR introduces a new `pkg/secrets` package providing authenticated encryption with compound key derivation for multi-tenant SaaS applications.

The package uses AES-256-GCM with HKDF-SHA256 to derive unique per-tenant keys from application and workspace keys, ensuring cryptographic tenant isolation. Identical plaintext encrypted for different workspaces produces entirely different ciphertext, preventing data leakage across tenant boundaries.

## Key Features

- **GenerateKey()**: Cryptographically secure random key generation (32 bytes/256 bits)
- **EncryptString/DecryptString**: Text encryption with automatic base64 encoding
- **EncryptBytes/DecryptBytes**: Binary data encryption with nonce prepending
- **ValidateKeys()**: Key validation to ensure 32-byte keys
- **HKDF-SHA256 Derivation**: Compound key model combining app and workspace keys
- **Security Properties**:
  - Authenticated encryption (AES-256-GCM provides confidentiality + integrity)
  - Unique random nonce per encryption operation
  - Sensitive key material zeroed after use
  - Ciphertext tampering detected during decryption
- **Comprehensive Error Handling**: Error joining for matching both operation and root cause

## Test Coverage

- Key generation and uniqueness
- Encrypt/decrypt round-trip for strings and binary data
- Invalid key handling
- Ciphertext validation (base64 decode, length, tampering)
- Tenant isolation: different workspace keys produce different ciphertext
- Cross-tenant decryption fails as expected
- All tests use `t.Parallel()` per project conventions

## Changes

- `pkg/secrets/doc.go`: Package documentation with usage examples
- `pkg/secrets/secrets.go`: Core encryption/decryption implementation
- `pkg/secrets/errors.go`: Sentinel error types
- `pkg/secrets/secrets_test.go`: Comprehensive test suite (331 lines)
- `go.mod`: Added `golang.org/x/crypto v0.47.0` dependency

## Test Plan

- [x] All tests pass: `make test`
- [x] No linting issues: `make lint`
- [x] Encryption/decryption works correctly
- [x] Tenant isolation verified (different keys produce different output)
- [x] Tampering detection works
- [x] Key validation enforced

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,0 +1,44 @@
+name: Claude Code Review
+
+on:
+  pull_request:
+    types: [opened, synchronize, ready_for_review, reopened]
+    # Optional: Only run on specific file changes
+    # paths:
+    #   - "src/**/*.ts"
+    #   - "src/**/*.tsx"
+    #   - "src/**/*.js"
+    #   - "src/**/*.jsx"
+
+jobs:
+  claude-review:
+    # Optional: Filter by PR author
+    # if: |
+    #   github.event.pull_request.user.login == 'external-contributor' ||
+    #   github.event.pull_request.user.login == 'new-developer' ||
+    #   github.event.pull_request.author_association == 'FIRST_TIME_CONTRIBUTOR'
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+      issues: read
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Run Claude Code Review
+        id: claude-review
+        uses: anthropics/claude-code-action@v1
+        with:
+          claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+          plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
+          plugins: 'code-review@claude-code-plugins'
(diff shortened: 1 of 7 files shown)


dmitrymomot:
@claude review pls

dmitrymomot:
Thanks for the review. Addressed the two real issues, skipped the three suggestions.

**Fixed:**

1. **`GenerateKey()` error semantics** — Added `ErrKeyGenerationFailed` sentinel. `GenerateKey()` now returns it instead of `ErrKeyDerivationFailed`, which is reserved for HKDF derivation in `deriveKey()`.

2. **`assert.NotEqual` → `require.NotEqual`** — Fixed both occurrences in `TestTenantIsolation` and `TestEncryptString_Uniqueness` to match project conventions.

**Skipped:**

3. **Key rotation / version byte** — Not in the roadmap spec. Adding a version prefix to ciphertext format is a breaking change to introduce preemptively. If rotation is needed later, it can be added as a separate concern (e.g., a `Vault` type that tries multiple key versions) without changing the stateless function signatures.

4. **Stronger key storage docs** — Current guidance matches the style of other `pkg/` packages (`jwt`, `cookie`). Opinionated production deployment advice belongs in consumer app docs, not a utility package.

5. **Benchmarks** — Comparable packages in the repo (`totp`, `jwt`) don't have benchmarks either. Can be added later if performance tracking becomes relevant.

See 5c8be7b.

dmitrymomot:
@claude review changes

PR #5, opened 2026-02-07
feat(pkg/token): add compact URL-safe signed token package

dmitrymomot:
## Summary

Introduces a new `pkg/token` package that provides compact, URL-safe signed tokens with truncated HMAC-SHA256 signatures. This package is designed for short-lived, single-use tokens commonly used in email verification links, password reset URLs, invite tokens, and magic links.

## Features

- **Generic functions** for any JSON-serializable payload type
- **Truncated HMAC-SHA256** signature (8 bytes / 64 bits) for compact size
- **Constant-time signature verification** to prevent timing attacks
- **URL-safe output** using base64url encoding (no padding, no special characters)
- **Zero external dependencies** (stdlib only)

## Token Format

Tokens consist of two base64url-encoded segments separated by a dot:
```
base64url(json(payload)).base64url(hmac-sha256[:8])
```

## Security Considerations

- 8-byte (64-bit) signature provides adequate brute-force resistance for short-lived tokens
- Designed to pair with server-side state (database flags, one-time nonces)
- Recommended for tokens with limited lifetime and single-use patterns
- Not suitable for long-lived bearer tokens (use JWT for those)

## Test Coverage

- 88.9% code coverage
- Comprehensive unit tests covering:
  - Round-trip token generation and parsing
  - Various payload types (structs, maps, primitives, slices)
  - Error cases (invalid tokens, tampered payloads, wrong secrets)
  - URL-safety verification (no +, /, or = characters)
  - Token determinism
- Performance benchmarks for both small and large payloads

## Test Plan

- All token package tests pass with 88.9% coverage
- No regression in other packages
- Linting passes successfully

PR #2, opened 2026-02-07
feat: add qrcode package for QR code generation

dmitrymomot:
## Summary
- Adds a new utility package for generating QR code images from string content
- Supports both raw PNG bytes and base64-encoded data URIs for HTML embedding
- Uses medium error correction (recovers up to 15% damage)
- Includes comprehensive tests with 94.1% coverage

## Implementation Details
- `Generate(content string, size ...int)` produces PNG bytes with optional custom size (default 256px)
- `GenerateBase64Image(content string, size ...int)` produces data URIs for direct HTML use
- Proper error handling with dedicated error types: `ErrEmptyContent`, `ErrFailedToGenerateQRCode`
- Well-documented for integration with the totp package for authenticator app setup

## Test plan
- [x] All qrcode package tests pass with 94.1% coverage
- [x] Tests verify PNG header validity
- [x] Tests cover edge cases: empty content, whitespace, custom sizes, defaults
- [x] Integration with totp package documented and verified in package examples

PR #6, opened 2026-02-07
Add ratelimit package with sliding window implementation

dmitrymomot:
## Summary

- Introduces a new `pkg/ratelimit` utility package for rate limiting with a sliding window algorithm
- Provides pluggable counter backends: in-memory and Redis-backed with automatic fallback
- Includes flexible key extraction strategies for various rate limit dimensions (IP, fingerprint, path, headers)
- Comprehensive test suite with 761 lines covering core functionality

## Key Features

- **Sliding Window Algorithm**: Smoothly blends request counts from previous and current windows to avoid boundary burst problems
- **Multiple Backends**:
  - `MemoryCounter`: In-process counter with background cleanup for single-process deployments
  - `RedisCounter`: Distributed counter with automatic fallback to memory on Redis failure
- **Key Extractors**: Pre-built extractors (IP, fingerprint, path, header) plus composite key support
- **Query Methods**: `Allow()` increments counter, `Peek()` checks without incrementing

## Test Plan

- All tests use `t.Parallel()` at function and subtest level following framework conventions
- Covers counter backends, key extraction, rate limit enforcement, and error cases
- Memory counter cleanup and Redis fallback scenarios tested


dmitrymomot:
@claude review pls

dmitrymomot:
Thanks for the thorough review. Pushed `84f7727` with the doc improvements. Here's the disposition of each item:

### Fixed (3 items)

**#2 — Default `"rl:"` prefix**: Added comment documenting the default namespace behavior.

**#4 — retryAfter guard**: Added comment explaining the floating-point rounding edge case.

**#7 — CleanupInterval zero-value**: Added doc comment to the struct field: zero = default (1 minute), negative = disabled.

### Not issues (9 items)

**#1 — "Race condition in MemoryCounter"**: False positive. Both `Increment` (line 62) and `Get` (line 79) acquire `m.mu.Lock()` *before* any access to `m.windows` or `w.expiresAt`. The `time.Now()` call touches no shared state. All shared memory is fully serialized by the mutex. The race detector confirms this — `go test -race` passes clean.

**#3 — Integer overflow**: `int64` holds 9.2×10¹⁸. Even at 1 billion requests/second sustained for 292 years you wouldn't overflow. Adding overflow guards would be dead code.

**#5 — "Use EXPIRE NX instead of EXPIRE"**: This would introduce a bug. If a key is first created late in a window cycle, the TTL is anchored to that creation time. With `NX`, subsequent increments in the next cycle inherit the old (nearly-expired) TTL, and the key vanishes mid-window. Refreshing TTL on every increment is intentionally correct — it ensures the key lives for at least `2*window` from the most recent write.

**#6 — Map growth**: Same characteristic as `pkg/cache/Memory`. The janitor bounds growth over time. A `MaxEntries` cap is a reasonable future enhancement but not a correctness issue for the initial implementation.

**#8 — RedisCounter concurrency test**: Redis `INCRBY` is atomic server-side. The fallback path uses `MemoryCounter`, which already has a concurrency test (`TestMemoryCounter_ConcurrentAccess`). A concurrent Redis test would be testing Redis itself, not our code.

**#9 — Boundary burst test**: The sliding window algorithm is validated by `TestLimiter_SlidingWindow/blends_previous_and_current_window`. A precise boundary test requires clock injection; without it the test would be timing-dependent and flaky.

**#10 — `_ context.Context` comment**: Standard Go idiom for interface implementations that don't use a parameter. No explanation needed.

**#11 — Error naming inconsistency**: Both `ErrRateLimited` (past participle = state) and `ErrInvalidLimit` (adjective = property) are idiomatic Go error names. Renaming would be churn with no clarity gain.

**#12 — Export `storageKey`**: Only used by `MemoryCounter`. Exporting adds public API surface for no consumer benefit.

PR #17, opened 2026-06-26
feat(request): HTTP request-reading helpers (input-side counterpart to render)

dmitrymomot:
## Summary

New standalone `request` package — the input-side counterpart to `render`. Stateless, **reflection-free** free functions that read data off an `*http.Request` into Go values. Stdlib only; flat layout; black-box tests.

Design spec and plan are included in the branch:
- `docs/superpowers/specs/2026-06-26-request-package-design.md`
- `docs/superpowers/plans/2026-06-26-request-package.md`

## What's included

- **Typed accessors** `Query` / `Path` / `Header` / `Cookie` / `FormValue`, each with `…Func` (custom-parser) variants, plus `…Slice` (repeated keys) and `…Split` (delimited) for query/form. One generic `parse[T]` engine: scalar type-switch → `time.Duration` → `encoding.TextUnmarshaler` (so `uuid.UUID`, `netip.Addr`, `time.Time`, custom enums work with **no `reflect`**).
- **Contract:** absent/empty ⇒ default or zero value + nil error; present-but-unparseable ⇒ `*request.Error{Kind: Malformed}`. Optional variadic default. `Has*` presence predicates close the zero-vs-absent gap.
- **Body decoding:** strict `DecodeJSON` (1 MiB cap → 413, `Content-Type` → 415, unknown-field/trailing/empty → 400) with `BodyOption` (`WithMaxBytes` / `AllowUnknownFields` / `SkipContentType`); `RawBody`; multipart `File` / `Files`; `IsContentType`.
- **`ClientIP`** — best-effort CDN/proxy header resolution (CF-Connecting-IP → True-Client-IP → Fastly → X-Real-IP → Forwarded → X-Forwarded-For → RemoteAddr), with `WithClientIPHeaders` (pin a header) and `WithTrustedProxies` (spoof-resistant right-to-left XFF walk). Default mode is documented as spoofable.
- **Pagination** — `QueryPage` (offset) / `QueryCursor` (opaque token) with `PageOption`; non-numeric → 400, out-of-range → clamped.
- **Errors:** one `*Error{Source, Key, Kind, Err}` + `StatusCode(err)` → 400/413/415. Options are split into three concern-scoped types (`BodyOption` / `ClientIPOption` / `PageOption`) so an irrelevant option is a compile error.

## Testing & quality

- Black-box tests (`package request_test`), `httptest`-driven; runnable godoc examples; a `parse` benchmark.
- `just check` (go fmt, goimports, betteralign, go vet, golangci-lint, nilaway, modernize, `go test -race`) is **green across the whole module**.
- Package coverage **84.6%**.
- Built task-by-task with a per-task review on each, plus a final whole-package review: **no Critical findings, no security issues**; all flagged items addressed.

## Non-goals (deliberate)

No validation-rules engine, no struct-tag binding/reflection, no content negotiation, no query/form merge, no signed cookies/CSRF/sessions.

### DIFF
--- a/docs/superpowers/plans/2026-06-26-request-package.md
+++ b/docs/superpowers/plans/2026-06-26-request-package.md
@@ -1,0 +1,2072 @@
+# request Package Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Build a standalone `request` package of stateless, reflection-free free functions that read data off an `*http.Request` into Go values — typed accessors, strict body decoding, and focused readers (BearerToken, ClientIP, pagination).
+
+**Architecture:** Every typed accessor funnels a raw string through one private generic engine (`parse[T]`: scalar type-switch → `time.Duration` → `encoding.TextUnmarshaler`) and a shared `resolve` helper that applies the missing/default/malformed contract. Per-source wrappers (`Query`, `Path`, `Header`, `Cookie`, `FormValue`) differ only in how they pull the raw string. Body decoding, ClientIP, and pagination each carry their own concern-scoped functional-option type. Failures are one typed `*Error{Source, Key, Kind, Err}`; `StatusCode` maps it to 400/413/415.
+
+**Tech Stack:** Go 1.26 standard library (`encoding`, `encoding/json`, `errors`, `fmt`, `mime`, `mime/multipart`, `net`, `net/http`, `net/netip`, `strconv`, `strings`, `time`); `testify` in tests only; `just` task runner.
+
+**Spec:** [docs/superpowers/specs/2026-06-26-request-package-design.md](../specs/2026-06-26-request-package-design.md)
+
+## Global Constraints
+
+- **Go 1.26**, module `github.com/dmitrymomot/forge`; new package import path `github.com/dmitrymomot/forge/request`.
+- **Stdlib only** in package code — no external dependencies. `testify` (already a module dep) allowed in tests only.
+- **Black-box tests ONLY** — every `*_test.go` declares `package request_test` and imports the package.
+- **Flat layout** — files live directly under `request/`, no subfolders.
+- **No reflection, no struct tags, no `init()`, no global mutable state** — interface assertions only (`encoding.TextUnmarshaler`).
+- **No builder pattern** — options are functional, split into three concern-scoped types: `BodyOption`, `ClientIPOption`, `PageOption`.
+- **Single-line errors**, prefixed `request:` and `%w`-wrapped; the wrapped cause stays reachable via `Unwrap`. No embedded stacks/blobs.
+- **Use `just` recipes** — `just test ./request/` to run; `just check` (fmt + lint + test) before finishing.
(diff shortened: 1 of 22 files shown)


dmitrymomot:
Follow-up `18c35ac` — addressed all remaining items from the per-task and final-package reviews:

**Polish**
- `Files` now returns `http.ErrMissingFile` for a missing key (matches `File`, `errors.Is`-friendly, no per-call allocation).
- `DecodeJSON` fetches `Content-Type` once.
- `BearerToken` godoc now documents the case-insensitive scheme + whitespace trimming.

**Coverage** (84.6% → 86.2%)
- `Error.Error()` no-cause path; `Kind.String()` all four kinds (prior commit).
- `BearerToken` case-insensitive scheme; `PathFunc`/`HeaderFunc`/`CookieFunc` smoke tests.
- `FormValue` malformed (→ typed `*Error`) and absent-with-default.
- `DecodeJSON(WithMaxBytes(0))` disables the cap.
- `ClientIP` with `WithTrustedProxies` ignores `Forwarded`; missing-file sentinel for `File`/`Files`.

`just check` (fmt, vet, golangci-lint, nilaway, betteralign, modernize, `go test -race`) green across the whole module.

Deliberately left as-is (not defects): the `parse[T]` "unsupported T" → `*Error{Malformed}` mapping (compile-time-only, design has no `KindUnsupported`, now covered by a test) and the silent `ParseForm` error in form accessors (matches the `query.go` convention).

dmitrymomot:
### Review feedback addressed (6b99b04)

- **Fixed — `resolveSplit` default fallback** (`parse.go`): `QuerySplit`/`FormSplit` on a value that is empty after splitting (e.g. `tags=,,`) now return the provided default, matching `resolveSlice`. Previously returned `nil`.
- **Fixed — missing-file semantics** (`errors.go`, `body.go`, spec): added `KindMissing`. `File`/`Files` report an absent file as `KindMissing` (not `KindMalformed`), still wrapping `http.ErrMissingFile`, so handlers can switch on `Kind` *or* `errors.Is(err, http.ErrMissingFile)`.
- **Fixed — `HasHeader` consistency** (`header.go`): a present-but-empty header (`X-Custom:`) now reports `true`, like `HasQuery`/`HasForm`.
- **Documented — deliberate `ParseForm` discard** (`form.go`): the form accessors are best-effort — a broken/absent body reads as no values, matching `net/http`'s own form accessors. Added inline comments.
- **Skipped — `r.URL.Query()` re-parse note** (`query.go`): caching would require per-request state, which breaks the stateless free-function design; stdlib doesn't cache it either (as the bot's own note acknowledges).

New tests cover each fix. `just check` (fmt, vet, golangci-lint, nilaway, betteralign, modernize, `go test -race`) is green across the whole module.

@claude ready for re-review.

dmitrymomot:
Follow-up: fixed a minor IPv6 issue in `ClientIP` — IPv4-mapped IPv6 addresses (`::ffff:a.b.c.d`, as a dual-stack listener reports IPv4 peers) are now `Unmap`-normalized to plain IPv4 in `validIP`/`remoteHost`. Without this they stringified in `::ffff:` form and, more importantly, a v4-mapped proxy hop did **not** match an IPv4 `WithTrustedProxies` prefix, so the spoof-resistant walk returned the proxy instead of the real client. Added IPv6 + v4-mapped test coverage. `just check` green.

PR #18, opened 2026-06-26
feat(htmx): re-add dropped conveniences + render.Components

dmitrymomot:
## Summary

Re-adds the useful conveniences the v1 `pkg/htmx` had and the v2 rewrite dropped, without reopening the rewrite's boundaries (`htmx` stays headers-only + stdlib-only; `render` stays HTMX-agnostic). Driven by a brainstorm → spec → plan → subagent-executed TDD flow.

New public surface:

**`htmx`**
- `type Swap string` — swap constants are now typed (`Reswap`, `LocationOptions.Swap` retyped); modifiers still read as `SwapInnerHTML + " swap:1s"`.
- `RedirectExternal(w, r, url, status...)` — explicit off-site full-page redirect (`HX-Redirect`, never `HX-Location`); no URL validation, documented for developer-controlled destinations only.
- `RedirectBack(w, r, fallback, status...)` / `RedirectBackParam(w, r, param, fallback, status...)` — return-redirect from a query param, **open-redirect-safe** (`safeLocalPath` honors only rooted same-origin paths; rejects `https://`, `//host`, `/\host`, etc.), HTMX-aware via `Redirect`.
- `LocationTarget(w, r, path, target, status...)` — path+target shortcut over `LocationWith` (terminal, no error — a two-string payload can't fail to marshal).

**`render`**
- `Components(ctx, w, status, ...Component) error` — generic, transactional, **HTMX-agnostic** multi-fragment writer (mirrors `Templ`: pooled buffer, nothing written on error). Plus `ErrNoComponents`.

**Out-of-band swaps** are documented as a templ-composition pattern (markup carries `hx-swap-oob`, rendered via `render.Templ`/`render.Components`, with `Reswap(SwapNone)` for OOB-only) — no feature code, since OOB is a body/markup concern.

Intentionally **not** restored (both by design): the obsolete `Config`/`With*` aggregator (tied to the removed `internal/context.go` pipeline), and exported `HeaderHX*` constants (kept encapsulated).

## Design & plan
- Spec: `docs/superpowers/specs/2026-06-26-htmx-extend-conveniences-design.md`
- Plan: `docs/superpowers/plans/2026-06-26-htmx-extend-conveniences.md`

## Testing
- `just lint` — clean (`go vet`, `golangci-lint` 0 issues, `nilaway`, `betteralign`, `modernize`).
- `just test ./...` — green with `-race`; `htmx` 98.5%, `render` 97.8% coverage.
- Black-box tests only; runnable `ExampleRedirectBack` / `ExampleRedirectExternal`.
- Final whole-branch review: ready to merge (no Critical/Important; security tests strengthened).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/superpowers/plans/2026-06-26-htmx-extend-conveniences.md
+++ b/docs/superpowers/plans/2026-06-26-htmx-extend-conveniences.md
@@ -1,0 +1,789 @@
+# htmx Conveniences Extension Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Re-add the useful conveniences the v1 `pkg/htmx` package had and the v2 rewrite dropped, plus an explicit external-redirect helper and a generic multi-fragment renderer, without reopening the rewrite's boundaries.
+
+**Architecture:** Three new free functions and one new type land in the existing `htmx` package (`response.go`), which stays headers-only and stdlib-only. One generic helper (`Components`) and one sentinel (`ErrNoComponents`) land in the existing `render` package, which stays HTMX-agnostic. Out-of-band swaps are handled by documentation (templ composition + `render.Components`), not by new feature code. TDD throughout, one committed task per deliverable.
+
+**Tech Stack:** Go 1.26, standard library only (`encoding/json`, `fmt`, `net/http`, `net/url`, `strings`, `context`), testify in tests, `just` task runner.
+
+## Global Constraints
+
+- Module path: `github.com/dmitrymomot/forge`; Go 1.26.
+- `htmx` and `render` are **stdlib-only** — no external runtime dependencies (testify is test-only).
+- **Black-box tests only:** test files are `package htmx_test` / `package render_test` and exercise only the exported surface.
+- Errors are **single-line, package-prefixed, wrapped** (e.g. `errors.New("render: no components")`, `fmt.Errorf("render: render component: %w", err)`).
+- **No builder pattern, no global mutable state** — free functions and plain values.
+- Flat package layout; files split by responsibility.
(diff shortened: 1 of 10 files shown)


dmitrymomot:
### Review feedback addressed (1d947d9)

Thanks for the thorough pass. All eight comments handled:

- **RedirectExternal alias footgun** — godoc now states it's behaviorally identical to `Redirect` and exists only to make the off-site intent explicit at the call site / in review.
- **`url` param shadows `net/url`** — renamed the parameter to `rawURL` in `PushURL`, `ReplaceURL`, `Redirect`, `RedirectExternal`, and `Location` (the shadow appeared once `net/url` was imported for `safeLocalPath`).
- **`LocationTarget` silent `_ =`** — added an inline comment documenting the invariant (path+target are strings, so the payload always marshals).
- **`Swap` doc** — clarified that the no-conversion convenience applies only to *untyped* string constants; a `string` variable needs `SwapInnerHTML + Swap(modifier)`.
- **`RedirectBack` fallback** — godoc now states `fallback` is developer-controlled and is NOT run through the safe-path check, so a user-controlled value must never be passed.
- **`Components` transactional doc** — reworded to distinguish render-path (nothing written) from a post-`WriteHeader` write error (non-recoverable, returned for logging), matching `Templ`.
- **URL-encoded `//` test** — added `TestRedirectBackHonorsEncodedSlash`. One correction: `/%2F%2Fevil.com` is **honored as a safe same-origin path**, not a fallback case — `%2F` stays encoded in the path so the browser keeps it on the current origin (`url.Parse` → `Host=""`), and `http.Redirect` emits the value verbatim. Placing it in the fallback table (as suggested) would assert the wrong behavior, so the test asserts it is honored.

**Skipped (1):**
- **`t.Parallel()` in `components_test.go`** — the comment compared against `htmx`, but within the `render` package **0 of 11** test files use `t.Parallel()`. Adding it only here would make `components_test.go` the lone outlier; leaving it omitted keeps the package consistent. Happy to flip the whole `render` suite to parallel separately if preferred.

`just lint` clean (0 issues); `just test ./...` green with `-race` (htmx 98.5%, render 97.8%).

@claude ready for re-review.

PR #20, opened 2026-06-28
feat(crypto): P1 crypto primitives — clock, randx, subtlex, hashx, redact, kdf, keyset, sign, password, secret, token

dmitrymomot:
## P1 Crypto Primitives

Implements the P1 crypto layer per [docs/superpowers/plans/2026-06-28-crypto-primitives.md](docs/superpowers/plans/2026-06-28-crypto-primitives.md): **nine crypto packages plus two minimal P0 seams**, built stdlib-first with `golang.org/x/crypto` confined to `kdf`/`password`/`secret`. Executed task-by-task with TDD via subagent-driven-development; every package got a spec+quality review and a final whole-branch review.

### Packages (5 dependency waves)

| Wave | Packages |
|---|---|
| 0 | `clock` (testable time seam), `randx` (crypto/rand helpers) |
| 1 | `subtlex` (constant-time length-safe compare), `hashx` (SHA/HMAC/file digest), `redact` (`Secret[T]` + scrub), `kdf` (HKDF + Argon2id) |
| 2 | `keyset` (versioned keyring), `sign` (HMAC + rotation), `password` (Argon2id + bcrypt + needsRehash) |
| 3 | `secret` (AES-256-GCM / XChaCha20-Poly1305 AEAD box + rotation) |
| 4 | `token` (generic `Codec[T]` opaque signed/encrypted expiring tokens) |

### Design

- **Key wiring "Approach A":** `New(key)` for the single-key case, `FromKeyset(ks)` for rotation; the key version rides in the encoded output (`sign` → `"<ver>.<mac>"`, `secret` → `version-byte‖nonce‖ct+tag`, `token` → `body.ver.mac`).
- **Constant-time everywhere:** every secret/MAC/hash comparison routes through `subtlex` (length-safe double-HMAC); `secret`/`token` rely on AEAD tag authentication. No `==`/`bytes.Equal` on secrets.
- **No oracle:** `secret.Decrypt` and `token.Parse` collapse every failure to a single sentinel without leaking the cause; `token.Parse` verifies the signature *before* decoding attacker-controlled bytes.
- **Options, not builders;** `errors.Is` single-line sentinels; black-box tests only.
- `golang.org/x/crypto` promoted indirect→direct (the only new direct dep), confined to `kdf`/`password`/`secret`.

### Verification

- `just check` fully green: gofmt/goimports/betteralign clean, **golangci-lint 0 issues**, nilaway/betteralign/modernize clean, **`go test -race` passes all packages**.
- New-package coverage: clock 100%, redact 100%, hashx 95%, keyset 94.2%, sign 91.3%, kdf 90.9%, password 87.3%, token 83.3%, randx 80%, secret 79.7%, subtlex 70% (sub-100 figures are unreachable-without-fault-injection RNG/AEAD/error branches).

### Review outcome

Final whole-branch review: **ready to merge**, no Critical/Important. It verified rotation correctness end-to-end, constant-time discipline, the no-oracle property, and `x/crypto` confinement. One substantive cross-package finding was applied: **`keyset` now enforces the 0..255 version range at construction** (`ErrBadKeyMaterial`) — previously a version ≥256 failed late at first `Encrypt` with a misleading error — since the wire format encodes the version in a single byte. Added doc notes on `keyset`'s returned-slice aliasing and `token`'s nonce intent (distinctness, not anti-forgery).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/clock/clock.go
+++ b/clock/clock.go
@@ -1,0 +1,49 @@
+package clock
+
+import (
+	"sync"
+	"time"
+)
+
+// Clock reports the current time. Production code depends on this interface rather
+// than calling time.Now directly, so time-dependent logic is deterministic in tests.
+type Clock interface {
+	Now() time.Time
+}
+
+type systemClock struct{}
+
+func (systemClock) Now() time.Time { return time.Now() }
+
+// System returns the production clock backed by time.Now.
+func System() Clock { return systemClock{} }
+
+// Mock is a goroutine-safe, controllable Clock for tests.
+type Mock struct {
+	t  time.Time
+	mu sync.Mutex
+}
+
+// NewMock returns a Mock fixed at t.
+func NewMock(t time.Time) *Mock { return &Mock{t: t} }
+
+// Now returns the mock's current time.
+func (m *Mock) Now() time.Time {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.t
+}
+
+// Set replaces the mock's current time.
+func (m *Mock) Set(t time.Time) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.t = t
+}
+
+// Advance moves the mock's current time forward by d.
+func (m *Mock) Advance(d time.Duration) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.t = m.t.Add(d)
+}
--- a/clock/clock_test.go
+++ b/clock/clock_test.go
@@ -39,5 +39,5 @@ func TestMock_Advance(t *testing.T) {
 
 func TestMock_ImplementsClock(t *testing.T) {
 	var _ clock.Clock = clock.NewMock(time.Now())
-	var _ = clock.System()
+	var _ clock.Clock = clock.System() //nolint:staticcheck // compile-time interface assertion; QF1011 inference suggestion does not apply
 }
--- a/clock/clock_test.go
+++ b/clock/clock_test.go
@@ -1,0 +1,43 @@
+package clock_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/dmitrymomot/forge/clock"
+)
+
+func TestSystem_NowIsCurrent(t *testing.T) {
+	before := time.Now()
+	got := clock.System().Now()
+	after := time.Now()
+	assert.False(t, got.Before(before))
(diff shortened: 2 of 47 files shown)


dmitrymomot:
### Review feedback addressed — `b8feb68`

Thanks for the thorough review. All seven findings handled below; `just check` is fully green (0 lint findings, all packages pass `-race`).

**Important**

1. **`redact.String` rune-safety** ✅ Fixed — now slices on `[]rune` boundaries instead of bytes, so multi-byte UTF-8 secrets no longer corrupt. Added a CJK test: `redact.String("一二三四五六七八九十")` → `"一二三四***七八九十"`.
2. **`secret.Encrypt` dead guard / wrong sentinel** ✅ Fixed — removed the `b.ver < 0 || b.ver > 255` guard entirely. It became provably unreachable once `keyset` started enforcing `0..255` at construction (single-key boxes are always version 0), and the `ErrInvalidKeySize` it returned was the wrong sentinel anyway.

**Minor**

3. **`errIsBcryptMismatch` → `errors.Is`** ✅ Fixed — now `errors.Is(e, bcrypt.ErrMismatchedHashAndPassword)`.
4. **`hashx.SHA512Base64` missing** ✅ Fixed — added for API symmetry with `SHA256Base64`, plus a test deriving the expected value from the SHA-512 known-answer digest.
5. **`keyset.All()` per-call alloc/sort** ✅ Fixed — `Keyset` is immutable, so the descending version order is now computed once in `New` (stored in `sortedVersions`) and `All()` iterates it with zero per-call allocation.
6. **`WithBase64Keys` fail-fast** ✅ Fixed — now accumulates *every* malformed pair into `c.errs` (`continue` instead of `return`), consistent with `WithPrimary`/`WithRetired`, so a bad env value reports all bad entries at once.
7. **Unreturned sentinels `sign.ErrBadSignature` / `password.ErrMismatch`** — **kept, by decision.** These are a deliberate, doc-commented forward-API: `sign.Verify`/`VerifyString` are boolean and `password.Verify` reports a wrong password via `ok=false` with a `nil` error (a wrong password is intentionally *not* an error — returning `ErrMismatch` there would break that contract and leak validity as an error). They exist as canonical sentinels for consumers wrapping these results in typed errors. Happy to drop both exports if you'd prefer the comment-only convention removed instead — just say so.

@claude ready for re-review.

PR #29, opened 2026-07-04
feat(web/subroute): chi.Mount-style prefix mounting for std ServeMux

dmitrymomot:
## Summary

New `web/subroute` package: mount any `http.Handler` under a path prefix on a standard `*http.ServeMux`, chi.Mount-style — plus a path-value fallback seam in `web/request` so `request.Path` keeps working across mount boundaries.

- **`subroute.Mount(mux, prefix, h)`** registers the prefix (exact + subtree) and dispatches with the prefix stripped, so sub-apps are written root-relative. A bare-prefix request reaches the handler as `GET /` (plain `http.StripPrefix` can't do this — it leaves an empty path that a nested ServeMux 301-redirects).
- **Wildcard prefixes** (`/app/{tenant}/dashboard`): prefix params are captured at the mount boundary because the std router discards a parent mux's path values when a nested mux matches. Stripping is segment-based on the escaped path (`%2F`-safe), with the `Path`/`RawPath` invariant preserved.
- **`request.WithPathValues(ctx, vals)`** is the new seam: `Path`/`PathFunc`/`HasPath` now read `r.PathValue` first, then the fallback — so `request.Path[string](r, "tenant")` works (typed) inside mounted handlers, and nested mounts merge later-wins with no logic in subroute. `web/request` stays stdlib-only.
- Static-prefix mounts skip capture entirely (no context allocation, asserted by test). The caller's request is never mutated. Setup-time panics only; no runtime errors.
- Docs: full godoc with the middleware-composition patterns and the honest chi difference (mounted middleware sees the stripped path); runnable examples; benchmarks with a no-mount baseline (+5 allocs/op static, +10 wildcard, dominated by the inner mux re-dispatch).

Design spec and implementation plan are included under `docs/superpowers/`.

## Test plan

- Full `just test` green (race + cover): `web/subroute` 96.2%, `web/request` 85.9%
- Black-box tests cover: bare/subtree/wildcard/nested stripping, innermost-wins shadowing, current-mux precedence over captures, encoded `%2F` segments, query preservation, sibling-route fall-through, Go 1.22 method patterns inside mounts, middleware path visibility on both sides of the boundary, caller-request immutability, static no-capture, 13 panic cases
- `ExampleMount` / `ExampleMount_wildcardPrefix` run as tests with exact output
- Scoped lint clean (`golangci-lint`, `nilaway`, `betteralign`, `modernize`)

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -121,7 +121,7 @@ forge/                              # single go.mod at the root
 │
 ├── web/           # HTTP in and out: transport, responses, boundary security, client
 │   # shipped: httpserver hostrouter middleware request clientip problem
-│   #          recoverer reqlog requestid render htmx
+│   #          recoverer reqlog requestid render htmx subroute
 │   # planned: httpclient cookie csrf secheaders cors timeout compress
 │   #          assets idempotency iplist captcha (captcha/turnstile ...)
 │   #          webhookverify autocert
@@ -258,7 +258,7 @@ Shipped: `backoff retry singleflight parallel circuitbreaker cache (cache/redis)
 ### web/
 
 Shipped: `httpserver hostrouter middleware request clientip problem recoverer
-reqlog requestid render htmx`.
+reqlog requestid render htmx subroute`.
 
 - **`httpclient`** — core, build early. Resilient outbound `*http.Client`
   via a RoundTripper stack: per-attempt timeout, jittered retry, circuit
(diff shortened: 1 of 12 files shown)


dmitrymomot:
Addressed the review's minor observation: added a doc comment on `pathValue` (web/request/pathvalues.go) explaining why an empty `r.PathValue` result is a safe absence sentinel — ServeMux single-segment wildcards never match an empty segment. Observations 2 and 3 need no change (docs volume is repo convention; CI is authoritative for the suite runs).

PR #52, opened 2026-07-13
feat(rng): add gaming/rng provably-fair game-mechanics randomness + pgstore

dmitrymomot:
## gaming/rng — deterministic game-mechanics randomness

Ships `gaming/rng`: deterministic random outcomes for game mechanics over a frozen `rng/v1` derivation spec, with a casual CSPRNG entry point and a provably-fair commit–reveal seed-chain `Manager`, plus the `gaming/rng/pgstore` pgx driver.

### Architecture (three layers, one package)
- **`Stream`** — HMAC-SHA256 block expansion of `(serverSeed, clientSeed, nonce)`; rejection-sampled ints (no modulo bias), 53-bit floats, spec'd Fisher–Yates. Output is frozen forever under `Algorithm = "rng/v1"` and pinned by golden vectors — reproducible byte-for-byte in any language, so outcomes are third-party verifiable.
- **Mechanics** (pure over a `Stream`) — weighted `Table[T]` with audit `Version()` hash and pity (guaranteed-drop counter returned to the caller, never stored); `Ints` (multi-value per nonce, e.g. slot reels); `Card`/`NewDeck`/generic `Deal` (cards, raffles); dice.
- **`Manager`** (stateful) — commit–reveal lifecycle, atomic per-round nonce consumption, rotation-with-reveal, fail-closed `WithScope` tenancy, behind a 5-method `Store` seam with an in-package memory store and an isolated Postgres driver.

### Two entry points, same math
- `Casual()` — CSPRNG-seeded, zero ceremony (SaaS lootboxes, marketing wheels).
- `NewManager(store, …)` — provably fair: published SHA-256 commitment, player client seed, atomic per-bet nonce, `Rotate` reveals the server seed for verification.

### Provably-fair guarantees (adversarially reviewed)
- The active server seed is **never** exposed before reveal (`Seed.ServerSeed` gated on `StatusRevealed`; `Proof` has no server-seed field).
- A revealed pair can never be replayed (reveal atomically frees the active slot).
- Fail-closed multi-tenancy: a scope-hook error or empty scope fails every method with `ErrNoScope`; scope is an opaque equality key with no empty-string widening. Single-tenant apps pay zero ceremony.

### Postgres driver
`pgstore.New(pool)` implements `rng.Store` over pgx with an embedded goose migration (`forge_rng_seeds`, partial unique index on active `(scope, player_id)`). `ConsumeNonce` is one atomic `UPDATE … RETURNING nonce - 1`; `23505` → `ErrExists` for both PK and active-index collisions; `Reveal` is idempotent via `COALESCE`.

### Testing & performance
- Black-box tests throughout; golden vectors + an **independent** reference reimplementation of rng/v1 freeze the spec.
- pgstore ran **live against Postgres 16** (7/7, incl. concurrent-consume and a full play→rotate→verify end-to-end).
- Fuzz: 8.66M execs, 0 crashers. `gaming/rng` coverage **90.7%**. Concurrency tests under `-race` for both stores and `Manager.Play`.
- Zero-alloc hot path (gated): `Uint64`/`IntN`/`Float64` = 0 allocs/op. Measured optimization: `Uint64` reads directly from the block buffer, **−5.63%** (p=0.000, n=10; 26.92ns→25.41ns), byte-identical output (verified via the block-alignment invariant + golden vectors).
- `just lint` clean repo-wide (modernize/betteralign/nilaway/errcheck).

### Anti-scope
No game math (paylines/RTP), no certified-RNG claims, no bet/wallet integration (compose with `finance/ledger` for money).

### Review
Executed via subagent-driven development (implementer + spec/quality review per task; Manager and the perf pass reviewed on the most capable model). Final whole-branch adversarial review: **ready to merge**, 0 Critical / 0 Important; 3 Minor (1 doc fix applied; 1 plan-mandated `ErrStore`+`ErrExists` join that remains `errors.Is`-matchable; 1 theoretical 32-bit-only cast, N/A on forge's 64-bit target).

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -439,16 +439,6 @@ pricing logic (the billing anti-scope stands).
 
 Deps: `core/money`; `core/fsm` (planned).
 
-## gaming/
-
----
-
-**gaming/rng**
-
-Deterministic random outcomes for game mechanics over one explicitly specified, frozen derivation algorithm (`rng/v1`: HMAC-SHA256 block expansion of serverSeed/clientSeed/nonce; rejection-sampled ints, 53-bit floats, spec'd Fisher–Yates) — reproducible in any language, so outcomes are third-party verifiable forever. Two entry points, same `Stream` type: `Casual()` (CSPRNG, zero ceremony — SaaS lootboxes, marketing wheels) and a provably-fair seed-chain `Manager` (commit–reveal: published SHA-256 commitment, player client seed, atomic per-bet nonce, rotation with reveal) behind a 5-method `Store` seam — memory store in-package, `rng/pgstore` pgx driver with embedded migrations. Mechanics are pure functions over a `Stream`: uniform ints/floats, multi-value per nonce (slot reels), weighted `Table[T]` with audit `Version()` hash and pure pity (guaranteed-drop counter returned to the caller, never stored), shuffle/`Deal` (cards, raffles), dice. Fail-closed `WithScope` tenancy. Anti-scope: no game math (paylines/RTP), no certified-RNG claims, no bet/wallet integration (compose with `ledger`).
-
-Deps: `core/random`, `core/clock`, `core/id`.
-
 ## async/
 
 ---
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -439,6 +439,16 @@ pricing logic (the billing anti-scope stands).
 
 Deps: `core/money`; `core/fsm` (planned).
 
+## gaming/
+
+---
+
+**gaming/rng**
+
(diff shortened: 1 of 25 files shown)


dmitrymomot:
Addressed the review in `0a56c92`:

- **Doc gap (observation 2):** `SetClientSeed` and `Rotate` godoc now surface the concurrent-writer retry contract — a racing writer returns an error matching both `ErrStore` and `ErrExists`, and the caller should retry. Previously this was only in a code comment.

- **Package LOC (observation 1):** left as-is by design — the `Stream` → mechanics → `Manager` layers share the frozen `rng/v1` derivation spec and shouldn't be split across import boundaries (a verifier must replay the exact same derivation the Manager produced). Documented rationale in the PR body; the CI review concurred it's a reasonable call.

Both already-flagged theoretical items (the intentional `ErrStore`+`ErrExists` join, the 32-bit-only `int` cast) confirmed non-issues. `just lint` clean, all `gaming/rng` tests green.

dmitrymomot:
Addressed the remaining review nits in `1eee005`:

- **`Table.Pick` 32-bit int cast (M3):** hardened. `NewTable` now bounds the weight sum to `math.MaxInt` (platform int) instead of `math.MaxInt64`, so `Pick`'s `int(totalWeight)` cast — the value it hands to `IntN(int)` — is exact on 32-bit targets too, not just 64-bit. Added a comment at the cast site and renamed the validation case to `exceeds int`. On this repo's 64-bit target the behavior is identical (`MaxInt == MaxInt64`); the change removes the theoretical 32-bit truncation. Verified no other `uint64→int` cast in the package can overflow (`v % n64` is bounded by the caller's `int n`; suit/rank casts are `uint8`/[0,12]).

- **Package LOC (observation 1):** left as one package by design, and the measurement supports it — excluding comments/blank lines the package is **709 lines of actual code** (9 files), within `docs/design.md`'s `~250–850` single-responsibility guideline. The ~1016 figure counts the large normative `rng/v1` spec + godoc in `doc.go`, which is documentation, not complexity. Splitting `Stream`/mechanics/`Manager` across import boundaries would break the shared-derivation invariant (a verifier must replay the exact derivation the Manager produced), so no split.

`just lint` clean, `gaming/rng` 90.7% coverage, all tests green under `-race`.

PR #53, opened 2026-07-13
feat(lockout): brute-force lockout with escalating windows

dmitrymomot:
## auth/lockout

Login/OTP brute-force protection: per-identity authentication-failure counting with escalating lockout windows, composed over two existing store seams — `resilience/ratelimit.Store` (atomic `Incr` counter, TTL = failure-memory window) for failure counts and `resilience/cache.Store` (TTL-KV with `SetNX`) for lock markers. No new drivers ship here; both seams bring their own backends. Not rate shaping (that's `ratelimit`) and not cumulative caps (that's `quota`) — failure-triggered escalation only.

### API

- `New(counters, locks, opts...)` — options for threshold / baseLock / factor / maxLock / window / clock / scope. Defaults: 5 free failures, then 1m locks doubling to a 15m cap, failures remembered 30m.
- `Allow(ctx, key) (Result, error)` — read-only pre-check.
- `Fail(ctx, key) (Result, error)` — records a failure; crossing the threshold creates a lock whose duration escalates `min(base × factor^k, maxLock)`. Exactly one concurrent threshold-crosser creates the marker (`SetNX`); losers report the winner's expiry. A `Fail` while already locked increments the counter (escalating the next lock) but never extends the current one.
- `Reset(ctx, key) error` — clears counter and lock on success.
- `Do(ctx, key, fn) error` — wraps an attempt; only errors matching `ErrFailedAttempt` count, so infrastructure failures can never lock a user out. Returns a `*LockedError` when locked on entry or when a counted failure crosses the threshold.
- `Middleware(KeyFunc, opts...)` — net/http gate over the `Allow` half; locked → 429 + `Retry-After`, unlocked → next with the extracted key stashed in context. Fails closed (503) on store errors by default; `WithFailOpen()` restores availability-first.

Multi-tenant apps set `WithScope` to derive a tenant scope from the context on every call (fails closed with `ErrScope` on error or empty scope); single-tenant apps omit it and pay zero ceremony. Caller keys are SHA-256 hashed into store keys — PII hygiene, not secrecy.

### Verification

- `just test ./auth/lockout/...` — PASS, race-clean, **92.7%** coverage.
- `just lint` — **0 issues** (vet, golangci-lint, nilaway, betteralign, modernize).
- `FuzzKeyComposition` — 8.5M execs, 0 crashers (key-shape + non-collision invariants across unicode / huge / empty / colon keys).

### Benchmarks (darwin/arm64, Apple M3 Max)

| Benchmark | ns/op | B/op | allocs/op |
|---|---|---|---|
| AllowUnlocked (hot path) | 241 | 96 | 2 |
| AllowLocked | 314 | 112 | 3 |
| FailBelowThreshold | 341 | 96 | 2 |
| FailWhileLocked | 729 | 176 | 7 |

No optimization pass needed — profiling confirmed `keys()` is already at the irreducible 2-allocation floor (the two output strings; SHA-256/hex are inlined and non-escaping).

### Adversarial security review

A whole-branch review traced seven bypass/lockout invariants against the real code and the two consumed seams — all hold: single-lock `SetNX` race, fail-while-locked no-extend + escalation-advance, overflow-safe `lockDuration` clamp, `Do` infra-error-never-counts, tenancy fail-closed on Allow/Fail/Reset, middleware fail-closed 503, and disjoint counter/lock key isolation. Findings it surfaced are fixed in this branch: a `NaN` factor that slipped the `factor < 1` guard (and silently disabled escalated locks), a `Fail.Locked`-vs-`Allow` doc clarification, a hard `window >= maxLock` validation guard, and single-line rendering of the rare double-store-failure errors.

### DIFF
--- a/auth/lockout/bench_test.go
+++ b/auth/lockout/bench_test.go
@@ -1,0 +1,82 @@
+package lockout_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/dmitrymomot/forge/auth/lockout"
+	"github.com/dmitrymomot/forge/resilience/cache"
+	"github.com/dmitrymomot/forge/resilience/ratelimit"
+)
+
+func benchLocker(b *testing.B, opts ...lockout.Option) *lockout.Locker {
+	b.Helper()
+	counters := ratelimit.NewMemoryStore()
+	b.Cleanup(func() { _ = counters.Close() })
+	locks := cache.NewMemoryStore()
+	b.Cleanup(func() { _ = locks.Close() })
+	lk, err := lockout.New(counters, locks, opts...)
+	if err != nil {
+		b.Fatal(err)
+	}
+	return lk
+}
+
+// The hot path: every login attempt pays one Allow.
+func BenchmarkAllowUnlocked(b *testing.B) {
+	lk := benchLocker(b)
+	ctx := context.Background()
+	b.ReportAllocs()
+	for b.Loop() {
+		if _, err := lk.Allow(ctx, "user@example.com"); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+func BenchmarkAllowLocked(b *testing.B) {
+	lk := benchLocker(b, lockout.WithThreshold(1), lockout.WithBaseLock(time.Hour), lockout.WithMaxLock(time.Hour))
+	ctx := context.Background()
+	if _, err := lk.Fail(ctx, "u"); err != nil {
+		b.Fatal(err)
+	}
+	if _, err := lk.Fail(ctx, "u"); err != nil {
+		b.Fatal(err)
+	}
+	b.ReportAllocs()
+	for b.Loop() {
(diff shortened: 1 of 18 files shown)


dmitrymomot:
Addressed the two non-blocking doc observations from the review in bda1518 (doc-only, no behavior change):

1. **`Do` error interpretation** — documented in both `doc.go` and the `Do` method comment that a non-nil return is not always a denial (a post-success `Reset` store hiccup surfaces as an `ErrStore`-wrapped error even though credentials were valid), and that callers should decide rejection by matching `*LockedError` / `ErrFailedAttempt` rather than a bare `err != nil`.

2. **`Middleware` empty-key skip** — added a callout in the `doc.go` usage example that a `KeyFunc` reading attacker-controlled input must not return `""` unless the handler itself rejects the empty case.

CI (test, lint, claude-review) was green on the prior commit; the `docs/packages.md` union-of-deletions conflict with the merged totp entry was resolved in the merge (all shipped entries removed).

PR #64, opened 2026-07-17
feat(smartlink): web/smartlink destination-decision engine

dmitrymomot:
## Summary

Implements `web/smartlink` per the catalog entry — the destination-decision engine (TDS/smartlink core) — and removes its roadmap entry. Design spec: `docs/superpowers/specs/2026-07-17-web-smartlink-design.md`.

- **Two-phase API**: `Compile(Spec, ...Option) (*Link, error)` validates consumer-hydrated rule data fail-fast; `Link.Decide(Visit) Decision` is the infallible per-click hot path (every failure mode is a compile error). No net/http import — the caller builds the `Visit` from its own request facts and emits the `Decision` as the click event.
- **Closed typed matcher vocabulary** (sealed `Matcher` interface, no DSL): `Geo`, `Device`, `Locale`, `ParamEquals`, `TimeWindow`, `Percent`. Ordered rules, AND within a rule, first match wins, mandatory default target.
- **Deterministic bucketing, never RNG**: `Percent` gates and weighted splits hash (rule-name salt + `Visit.StickyKey`) via inlined FNV-1a (featureflag precedent), with independent salts so a gate and its split don't correlate. Empty sticky key fails closed (`Percent` no-match, splits take the first target).
- **URL templates**: `{country}`, `{device}`, `{locale}`, `{param.NAME}` parsed at compile — unknown macro is a construction error, never an empty substitution at fire time (postback precedent). Escaping is positional (path vs query), so macro values can't alter validated URL structure and `Decide` needs no error path.
- **`ParamPolicy`**: `ParamsDrop` (default) / `ParamsFill` / `ParamsOverride` merge of visit params into the final URL.
- **`core/clock` is the only dep**; `WithClock` for tests, `Visit.At` override for click-log replay. Tenancy = passed value (per-tenant rule sets → per-tenant compiled Links); no stored state to scope.

## Benchmarks (Apple M3 Max)

Post-benchmark optimization: the internal compiled matcher is a tagged union instead of an interface, which devirtualizes the match loop and keeps the `Visit` on the stack.

| bench | before | after |
|---|---|---|
| DecideLiteral | 53.55 ns/op, 128 B, 2 allocs | **23.47 ns/op, 0 B, 0 allocs** |
| DecideSplit | 47.80 ns/op, 128 B, 2 allocs | **15.73 ns/op, 0 B, 0 allocs** |
| DecideMacro | 85.20 ns/op, 208 B, 3 allocs | **55.85 ns/op, 80 B, 1 alloc** (the result string) |
| DecideMerge | 560.6 ns/op, 888 B, 13 allocs | **532.7 ns/op, 760 B, 11 allocs** |

## Testing

- Black-box suite (`smartlink_test`): compile validation table (every sentinel), matcher semantics (geo case-fold, locale primary-subtag rules, case-sensitive params, time-window bounds incl. exclusive `Until` and `Visit.At` replay), determinism + distribution checks for `Percent` and splits (5000-key tolerance bands), gate/split decorrelation, macro rendering + positional escaping, all three param policies, concurrent `Decide` under `-race`.
- `go test -race -cover ./web/smartlink/` → 95.9% (uncovered: two documented-unreachable defensive branches + the godoc Example).
- `just lint` clean.


dmitrymomot:
Addressed post-review: an independent review pass caught that authority-position macros (e.g. `https://cdn-{param.region}.example.com/`) rendered with `url.PathEscape`, which leaves `@` and `:` unescaped — a hostile param value could inject userinfo or a port into the host, contradicting the documented "macros cannot alter URL structure" invariant. Fixed in 987e6ab: escaping mode is now derived per macro from its elided prefix (authority / path / query). Authority values render only when entirely hostname-safe bytes and fail closed to empty otherwise (net/url forbids percent-escapes in a host, so encoding wasn't an option); path values additionally encode `:` to block scheme injection in relative templates. Added structure-safety tests for both, plus the sorted-query doc note on `ParamPolicy` from the review's minor observations. Benchmarks unchanged (hot path still 0 allocs).

dmitrymomot:
Superseded: the engine landed verbatim as the seed commits of #68 (first commit is a byte-identical checkout of this branch's web/smartlink, verified by empty diff), with the compiled artifact renamed Link→Compiled per the unified design (docs/superpowers/specs/2026-07-17-smartlink-unified-design.md). All review hardening from this PR is preserved.

PR #66, opened 2026-07-17
feat(attribution): web/attribution — marketing-touch capture package

dmitrymomot:
## Summary

New `web/attribution` package: captures marketing touches — `utm_*` params, ad-platform click IDs, affiliate sub-IDs — into a signed cookie and hands them back at conversion time. Answers "where did this signup come from" with exactly one stored touch per visitor; multi-touch models stay out.

- **Middleware** watches every request; when the query carries any configured param, the set is stored as a `Touch` in a signed cookie (`cookie.SetSigned`, tamper-proof). Capture is best-effort and never fails the request.
- **Policy**: `LastTouch` (default) overwrites on every new campaign visit; `FirstTouch` keeps the original until it expires. Expired first touches are replaced.
- **Attribution window** (default 30 days) bounds both the cookie `Max-Age` and a server-side timestamp check inside the signed payload, so a replayed or clock-gamed cookie past the window never counts.
- **`Touch(r)`** returns the stored touch at conversion (context first for same-request capture, verified cookie otherwise); **`Clear(w)`** expires the cookie so one touch isn't attributed twice.
- **`Pixel()`** serves the canonical 43-byte 1×1 transparent GIF with full no-cache headers (validated by `image/gif` decode in tests), capturing params from the pixel URL.
- Default param set: `utm_*` (6) + click IDs for Google (`gclid`/`gbraid`/`wbraid`/`dclid`), Microsoft, Meta, TikTok, X, LinkedIn, Pinterest, Reddit; `WithParams` replaces, `WithExtraParams` appends (sub-IDs).
- Hygiene: first value wins on repeated params, empty values skipped, values >512 bytes dropped (so junk can't crowd the 4 KiB cookie limit out from under real click IDs).
- Default cookie `__Host-attribution`, falling back to `attribution` when the codec policy can't satisfy `__Host-` (csrf precedent).
- Removes the shipped entry from `docs/packages.md` (catalog lists unbuilt packages only).

Tenancy note: like `csrf`/`cookie`, the touch cookie is inherently browser+domain scoped, so no `WithScope` seam applies (same precedent as the other web boundary packages).

## Benchmarks (Apple M3 Max)

```
BenchmarkMiddleware_NoQuery-14         192501620    6.196 ns/op       0 B/op    0 allocs/op
BenchmarkMiddleware_UnrelatedQuery-14    3237378    369.1 ns/op     448 B/op    5 allocs/op
BenchmarkMiddleware_Capture-14            382160     3165 ns/op    5259 B/op   47 allocs/op
BenchmarkTouch_CookieRead-14              471522     2588 ns/op    3016 B/op   45 allocs/op
BenchmarkPixel-14                         457978     2559 ns/op    4587 B/op   44 allocs/op
```

The hot path (no query string — the overwhelming majority of requests) is 0 allocs / ~6 ns. Capture and conversion reads run once per campaign landing / signup, dominated by HMAC verify + JSON — no perf-motivated complexity warranted (numbers above are the post-benchmark pass; the fast-path `RawQuery == ""` early return is the one measured win, taking the common case from ~370 ns/5 allocs to 6 ns/0 allocs).

## Test plan

- [x] `go test -race -cover ./web/attribution/...` — 93.6% coverage, black-box only
- [x] Capture→conversion round-trip, last-touch overwrite (full replace, no merge), first-touch keep (no cookie rewrite), expired-first-touch recapture
- [x] Window expiry server-side check, tampered-cookie rejection, oversized-touch best-effort drop (request still succeeds)
- [x] Pixel: valid 1×1 GIF (decoded), no-cache headers, HEAD omits body, params captured from pixel URL
- [x] `just lint` clean (vet, golangci-lint, nilaway, betteralign, modernize, integration tier)

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -73,20 +73,6 @@ filtering stay consumer-side.
 
 Deps: `core/clock`.
 
----
-
-**web/attribution**
-
-Marketing-touch capture: middleware records a configured param set
-(`utm_*`, click IDs, sub-IDs) into a signed cookie (`cookie` + `sign`)
-under first-touch or last-touch policy with an attribution window, and
-hands the stored touch back at conversion time. Includes a
-tracking-pixel endpoint (correct 1×1 GIF + no-cache). The "where did
-this signup come from" answer every SaaS wants and every affiliate
-platform requires; multi-touch models stay out.
-
-Deps: `web/cookie`, `crypto/sign`.
-
 ## view/
 
 ---
--- a/web/attribution/attribution.go
+++ b/web/attribution/attribution.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"log/slog"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/dmitrymomot/forge/core/ctxkey"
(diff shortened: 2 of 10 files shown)


dmitrymomot:
Addressed all three review findings in 3408bea:

- **Incompatible custom `__Host-` name**: `New` now panics when a custom `__Host-` cookie name can't be satisfied by the codec policy (same wiring-bug reasoning as the nil-codec panic), instead of every best-effort capture failing silently at Debug level. The default name still auto-falls-back to `attribution`. Covered by `TestNewPanicsOnIncompatibleHostPrefixName`.
- **Param slice aliasing**: `WithParams` clones the incoming slice and `WithExtraParams` clips before appending, so no option composition can write into a caller-owned backing array. Covered by `TestParamOptionsDoNotAliasCallerSlice` (spare-capacity base slice, asserts the backing array is untouched).
- **Doc overclaim**: doc.go and the `Middleware` comment now distinguish the empty-query zero-alloc fast path from unrelated-query requests (one parse, no cookie work), matching the benchmarks.

PR #67, opened 2026-07-17
feat(tenant): add data/tenant multi-tenancy package

dmitrymomot:
## Summary

Ships the `data/tenant` catalog entry: forge's multi-tenancy package. Stdlib only.

- **Resolver chain** — `type Resolver func(*http.Request) (string, error)` with `("", nil)` = not resolved / try next. Shipped: `Subdomain(base)`, `Domain(lookup)` over the storage-agnostic `DomainLookup` seam (+ `StaticDomains` in-memory impl), `Header(name)`, `Cookie(name)`, `PathPrefix(prefix)`, and `Context()` (gives an upstream-stamped tenant — e.g. API-key-derived — an explicit precedence slot).
- **Middleware** — `Middleware(resolvers...)`: precedence-ordered, first non-empty ID wins and overrides any pre-existing ctx tenant; resolver error → generic 500; unresolved → request passes through unchanged (upstream tenant preserved, otherwise untenanted). `Require` guards mandatory-tenancy routes with 404. Both satisfy `web/middleware.Middleware` structurally, no import.
- **Transport-agnostic carrier** — `NewContext` / `FromContext` / `Scope`; queue handlers and cron jobs stamp/read tenancy without HTTP. `tenant.Scope` matches every forge `WithScope(func(ctx) (string, error))` option, so it plugs in directly: `apikey.WithScope(tenant.Scope)`.
- **ScopeClause** — explicit parameterized SQL fragment, visible at every query, never auto-injected: `ScopeClause(ctx, "tenant_id", "$2")` → `Clause{SQL: "tenant_id = $2", Arg: id}`. The placeholder is written verbatim (`$n` pgx, `?` sqlite/mysql/clickhouse) — no dialect enum. Fail-closed on missing tenant (`ErrNoTenant`); column and placeholder are validated byte-wise so misuse can't smuggle SQL into the fragment.
- Misconfigured constructors (empty base/name, nil lookup, invalid prefix, nil resolver) panic with `errors.Is`-matchable sentinels — the hostrouter precedent.
- Catalog entry removed from `docs/packages.md` per the unbuilt-only rule; design spec committed under `docs/superpowers/specs/`.

## Design notes

- Host normalization mirrors hostrouter's zero-alloc `normalizeHost` (port/IPv6 brackets/FQDN dot stripped, lowercased; sub-slices unless uppercase input). `Subdomain` resolves only an exact single label in front of the base — `notapp.example.com`, `app.example.com.evil.io`, nested `a.b.base`, and the bare base all miss.
- Resolvers return the raw identifier they saw; slug→canonical-ID mapping stays a consumer wrapper or lives inside `DomainLookup`. Reserved subdomains (www/api) are deliberately not special-cased.
- `Header`/`Cookie` values are client-controlled; documented for trusted-gateway/internal use only.

## Benchmarks (Apple M3 Max, post-optimization)

| Benchmark | ns/op | B/op | allocs/op |
|---|---|---|---|
| Subdomain (hit) | 27.9 | 0 | 0 |
| Subdomain (miss) | 23.2 | 0 | 0 |
| Header | 20.3 | 0 | 0 |
| PathPrefix | 2.7 | 0 | 0 |
| Domain (StaticDomains) | 39.5 | 0 | 0 |
| Middleware end-to-end | 163.5 | 384 | 3 |
| ScopeClause | 36.8 | 16 | 1 |
| FromContext | 3.3 | 0 | 0 |

Optimization pass: `Header` was 53.3 ns / 1 alloc — `Header.Get` re-canonicalized `X-Tenant-ID` (canonical form is `X-Tenant-Id`) per call; pre-canonicalizing at construction makes it 20.3 ns / 0 allocs. Middleware's 3 allocs are the unavoidable `context.WithValue` + request clone; ScopeClause's 1 alloc is the fragment string.

## Testing

- Black-box unit tests (98.4% coverage): carrier round-trip, fail-closed `Scope`, every resolver's hit/miss/error/panic cases incl. host edge cases (port, case, FQDN dot, IPv6, boundary confusion), middleware precedence/override/pass-through/500, `Require`, `StaticDomains` normalization, ScopeClause valid/invalid matrices (incl. injection probes).
- Fuzz: `FuzzSubdomain` (15s, 1.1M execs) and `FuzzPathPrefix` (15s, 4.3M execs) — no failures; properties assert resolved labels/segments are well-formed and actually present in the input.
- Independent code-review agent pass before PR: no Critical/Important findings; the three Minor findings (middleware doc overstating untenanted pass-through, spec fuzz-target drift, coverage pins) are fixed in the second commit.
- No integration tier needed — no DB/network.

### DIFF
--- a/data/tenant/bench_test.go
+++ b/data/tenant/bench_test.go
@@ -1,0 +1,99 @@
+package tenant_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/dmitrymomot/forge/data/tenant"
+)
+
+func BenchmarkSubdomain(b *testing.B) {
+	resolve := tenant.Subdomain("app.example.com")
+	r := newRequest("acme.app.example.com", "/")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := resolve(r); id == "" {
+			b.Fatal("expected resolution")
+		}
+	}
+}
+
+func BenchmarkSubdomainMiss(b *testing.B) {
+	resolve := tenant.Subdomain("app.example.com")
+	r := newRequest("other.example.org", "/")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := resolve(r); id != "" {
+			b.Fatal("unexpected resolution")
+		}
+	}
+}
+
+func BenchmarkHeader(b *testing.B) {
+	resolve := tenant.Header("X-Tenant-ID")
+	r := newRequest("example.com", "/")
+	r.Header.Set("X-Tenant-ID", "acme")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := resolve(r); id == "" {
+			b.Fatal("expected resolution")
+		}
+	}
+}
+
+func BenchmarkPathPrefix(b *testing.B) {
+	resolve := tenant.PathPrefix("/t")
+	r := newRequest("example.com", "/t/acme/dashboard")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := resolve(r); id == "" {
+			b.Fatal("expected resolution")
+		}
+	}
+}
+
+func BenchmarkDomainStatic(b *testing.B) {
+	resolve := tenant.Domain(tenant.StaticDomains(map[string]string{"shop.acme.com": "acme"}))
+	r := newRequest("shop.acme.com", "/")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := resolve(r); id == "" {
+			b.Fatal("expected resolution")
+		}
+	}
+}
+
+func BenchmarkMiddleware(b *testing.B) {
+	h := tenant.Middleware(
+		tenant.Domain(tenant.StaticDomains(map[string]string{"shop.acme.com": "acme"})),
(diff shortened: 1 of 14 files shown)


dmitrymomot:
API rework in 537cc70: resolvers now always yield the **canonical tenant ID** (uuid/ulid/short id), never an alias — a subdomain label or domain is only ever an input to resolution.

- `Subdomain(base, lookup SubdomainLookup)` — the extracted label is translated to a tenant ID through the new storage-agnostic `SubdomainLookup` seam (`StaticSubdomains` in-memory impl shipped). Reserved names (www/api) are now naturally handled by the lookup returning not-found.
- `ErrDomainNotFound` → `ErrTenantNotFound`, shared by both lookup seams and `Map` fns: alias maps to no tenant → chain continues.
- New `Map(r Resolver, fn)` combinator adds the same alias→ID translation to any resolver (e.g. slugs in `PathPrefix` URLs).
- `Header`/`Cookie`/`PathPrefix` documented as carrying the ID itself unless wrapped with `Map`.

Benchmarks unchanged in character: Subdomain hit 33.4 ns / 0 allocs (map lookup added), miss 21.4 ns / 0 allocs, middleware 163.7 ns / 3 allocs. Coverage 98.7%; fuzz re-run clean; spec updated.

PR #69, opened 2026-07-18
feat(i18n): core/i18n mechanism-only i18n package + core/i18n/cldr data sibling

dmitrymomot:
## Description

Adds `core/i18n`, a zero-external-dep internationalization package built as **pure mechanism that ships no language data**, plus `core/i18n/cldr`, an opt-in sibling carrying the real CLDR data (44 plural rules, 22 locale format specs). The compiler enforces the one-way boundary: `cldr` imports `i18n`, `i18n` never imports `cldr`, so locale policy can never leak into core.

### Why this shape

An i18n package that hard-codes a locale table becomes a maintenance sink and forces every app to carry data it may not use. Here the supported locale set is whatever the application's catalogs declare — any tag works (`vi` is in the testdata to prove it) — and CLDR plural rules and format specs are opt-in, wired explicitly via `WithPlural`/`WithFormat`. An app that wants none of it pays zero data cost; an app that wants Ukrainian wires `cldr.Uk`.

### What's included

- **Bundle** — catalogs loaded from an `fs.FS` (`{tag}/{namespace}.json`), compiled once at `New`; resolution is precomputed exact-tag → base-language → default, so every lookup is array indexing. Rendering is **total**: never returns an error, never panics (missing key → key echo + optional miss handler; unresolved context → default locale; zero `Locale` → default).
- **Three API layers** over one Bundle: package-level context one-liners (`i18n.T(ctx, …)`), `Localizer` values for `html/template`, and explicit-locale `Bundle` methods.
- **Formatting** — numbers, percent, currency (over `core/money`, respecting each currency's minor units), and date/time, all driven by a per-locale `FormatSpec` with a neutral `Invariant` fallback. Hot-path `Append*` variants are 0-alloc.
- **Plurals** — a built-in zero-one-many `DefaultRule` for unconfigured locales; real CLDR rules come from `cldr`. The plural rule used is the rule of the locale a message was *found* in, and a catalog's `zero` form wins at n=0 regardless of the wired rule.
- **HTTP** — Accept-Language negotiation (bounded, total on adversarial input) and middleware with a cookie/query/header resolver chain that sets `Content-Language` and appends to `Vary`.
- **`cldr`** — 44 languages' plural rules and 22 locales' format specs, differential-tested against `golang.org/x/text` (test-only dependency).

### Notes for reviewers

- **`i18n` deps:** stdlib + `core/money` + `core/ctxkey` only. The `vendor/golang.org/x/text/*` entries in `go list -deps` are the standard library's own copies (reached via `net/http` → `x/net/idna`), not the external module — the external `golang.org/x/text` module is a test-only dependency of `cldr`.
- **CLDR correctness:** the `cldr` rules were verified against `x/text` v0.38, but that version is stale/buggy in a few places, so several rules were additionally cross-checked against the authoritative Unicode CLDR source and corrected: Macedonian (`%100!=11` one-exception), Hebrew (round-tens `many` removed at CLDR v42), and zh-CN/ko-KR date layouts (unpadded, unlike Japanese). Divergences from `x/text` are recorded in a `staleExceptions` map with reasons and pinned by explicit known-value tests.
- **`FormatFrFR`** uses U+00A0 as the group separator per the design doc; current CLDR specifies U+202F (narrow no-break space). Documented as a deliberate deviation — happy to switch if preferred.
- Coverage: 96.5% (`core/i18n`) / 100% (`core/i18n/cldr`). Benchmarks and a before/after optimization pass are included (`slices.SortStableFunc` in Accept-Language negotiation: −12% time / −25% allocs).

## Checklist

- [x] `just lint` passes
- [x] `just test` passes
- [x] New code has tests


dmitrymomot:
### Changes pushed (8b1ac94)

Addressed all 10 review findings (7 correctness, 3 cleanup). Full suite passes with `-race`, `just lint` is clean, `AppendPercent` stays 0-alloc.

- **Middleware nil-fn panic**: skip nil-fn `Resolver`s (zero value / `NewResolver(h,nil)`) in the Vary + resolution loops.
- **ValidateKeys/plural echo**: `T`/`TK` now fall through to a plural's `other` form (mirrors `TN`), so a validated plural key degrades to real text, never an echo.
- **Partial FormatSpec → empty dates**: `resolveFormat` backfills empty layouts + `DecimalSep` from `Invariant` (`GroupSep` left as valid no-grouping).
- **Percent float artifact**: round `ratio*100` to a 6-digit grid (`0.29 → 29%`); still zero-alloc, NaN/±Inf unchanged.
- **He plural doc**: corrected — CLDR v42 removed Hebrew's `many`; the function was already one/two/other.
- **Cleanups**: accurate Accept-Language 32-cap comment; shared exact-tag→base-language walk for `Parse`/`locIdx`; shared `checkPluralCollisions` for `setPlural` + cross-source merges.

**Two evidence-backed deviations** (verified each locale against cldr-json `main` directly rather than the finding text):
- **CLDR padding**: unpadded the *day* for en-US, es-ES, es-AR, es-CO, pl, tr, ar-AE — but **es-CO/pl/tr keep a padded month** (`d/MM`, `d.MM`) and **cs-CZ was left unchanged** (its CLDR short date `dd.MM.yy` is already padded; the finding was incorrect for cs).
- **24h time padding** (es-ES/cs `H:mm`): **not fixable** — Go's layout has no unpadded-24h token (`15` is always two digits), so these render padded `15:04`, now documented alongside the Turkish-percent gap.
- **isPluralMap ambiguity**: no heuristic can separate a category-only-keyed namespace from a real plural (`{one, other}` is English's own shape; uk uses `{one,few,many}`), so fixed as a documented deterministic rule + non-category-key escape + the `ValidateKeys` safety net, with a regression test.

Regression tests added for each fix. @claude ready for re-review.

PR #72, opened 2026-07-18
feat(data/pagination): keyset cursor + offset page-window pagination

dmitrymomot:
Implements the `data/pagination` package from the catalog — two pagination models as composable pieces that emit `(sql, args)` and run no queries themselves (imports no driver).

## Keyset (cursor) pagination — APIs, feeds, infinite scroll
- **`Keyset`** builds the portable **OR-of-ANDs** WHERE comparison and the ORDER BY list, with a placeholder-dialect option:
  - `Dollar` (`$n`, pgx default) reuses each column's placeholder → `args` = one value per column.
  - `Question` (`?`, ClickHouse/MySQL/SQLite) is positional and can't reuse, so `args` repeats the equality-prefix values in exact `?` order.
  - A caller-chosen `start` index lets the fragment compose after a query's existing args (e.g. a `tenant.ScopeClause` at `$1`).
- Mixed ASC/DESC handled; a **backward** cursor flips both the WHERE operators and the ORDER BY directions.
- **Security:** columns are validated identifiers (same rule as `tenant.ScopeClause`); operators are hardcoded; cursor values only ever bind as args — no user input reaches SQL text.
- **`Codec`** encodes the boundary row's key values into an opaque `base64url(JSON)` cursor, optionally HMAC-signed via `crypto/sign` (tamper-evident, rotation-aware via `FromKeyset`). `int64` keys round-trip losslessly (beyond 2^53); a number beyond `float64` range fails closed as `ErrBadCursor`.
- **`NewPage[T]`** assembles fetched rows (`LIMIT size+1`) plus next/prev cursors into `Page[T]`, handling forward and backward paging (sentinel trim + reversal).

## Offset (numbered) pagination — server-rendered admin tables
- `PageCount` / `Offset` drive the `LIMIT`/`OFFSET` query.
- `NewWindow` builds the numbered nav bar: first/last always shown, a `Span` of pages around the current one, **ellipses collapsed to the digit for one-page gaps**, per-link query strings preserving the request's other params — without mutating the caller's `url.Values`.

## Testing & perf
- Black-box tests, **99.5% coverage**, a `Decode` fuzz (15s+ clean), and benchmarks. Exact-string assertions pin the SQL emission for both dialects.
- Benchmarks (Apple M3 Max):
  ```
  WhereDollar     141 ns/op    168 B/op    5 allocs/op
  WhereQuestion   127 ns/op    168 B/op    5 allocs/op
  OrderBy          54 ns/op     48 B/op    2 allocs/op
  Encode          194 ns/op    176 B/op    4 allocs/op
  Decode          699 ns/op   1192 B/op   18 allocs/op
  NewWindow       701 ns/op   1744 B/op   20 allocs/op
  ```
- Post-benchmark optimization pass (measured): `NewWindow` encoded the non-page query params **once** and appends `&page=N` per link instead of cloning + re-encoding the whole map per page number — **2577→701 ns/op, 5680→1744 B/op, 69→20 allocs/op** (~4×).

## Housekeeping
- Deletes the `data/pagination` entry from `docs/packages.md` (shipped-package rule) and untags the `ops/auditlog` dependency.

Deps: `crypto/sign` only.

### DIFF
--- a/data/pagination/bench_test.go
+++ b/data/pagination/bench_test.go
@@ -1,0 +1,79 @@
+package pagination_test
+
+import (
+	"net/url"
+	"testing"
+
+	"github.com/dmitrymomot/forge/data/pagination"
+)
+
+var (
+	benchKeyset = pagination.Keyset{
+		{Column: "created_at", Desc: true},
+		{Column: "id", Desc: true},
+	}
+	benchCursor = pagination.Cursor{Keys: []any{"2026-01-01T00:00:00Z", int64(4242)}}
+)
+
+func BenchmarkWhereDollar(b *testing.B) {
+	b.ReportAllocs()
+	for b.Loop() {
+		if _, err := benchKeyset.Where(benchCursor, pagination.Dollar, 1); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+func BenchmarkWhereQuestion(b *testing.B) {
+	b.ReportAllocs()
+	for b.Loop() {
+		if _, err := benchKeyset.Where(benchCursor, pagination.Question, 1); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+func BenchmarkOrderBy(b *testing.B) {
+	b.ReportAllocs()
+	for b.Loop() {
+		if _, err := benchKeyset.OrderBy(false); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+func BenchmarkEncode(b *testing.B) {
+	codec, err := pagination.NewCodec()
+	if err != nil {
+		b.Fatal(err)
+	}
+	b.ReportAllocs()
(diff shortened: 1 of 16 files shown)


dmitrymomot:
Addressed the review feedback:

- **`Keyset.Where` args over-allocation** (both reviews) — `004b4d9` sizes the `args` slice to `n` for `Dollar` and `n(n+1)/2` only for `Question`.
- **`NewPage` panic on non-positive `size`** — `71a2aa7` returns `ErrInvalidSize` for `size < 1` instead of panicking on `rows[:size]`. Since `size` is often a request-derived `?limit=`, this turns a potential DoS into a fail-closed error, matching `Where`'s input validation. Added a test.

Local `just lint` and `go test -race` clean; coverage 99.6%.

PR #74, opened 2026-07-18
fix(web/smartlink): bound coalesced waits by caller context, close tenancy and resolver safety gaps

dmitrymomot:
## Description

Fixes the six actionable issues from the post-merge review of web/smartlink (PR #68), covering cache resilience and API safety.

**P1 — cache-miss coalescing dropped request deadlines.** Both coalescing points (`Cache.Get` for ref compiles, `Manager.lookup` for the Link read-through cache) blocked waiters unconditionally, so a stuck Store/Ref loader pinned every redirect request for that key indefinitely. The shared load now runs in one detached goroutine per key and every caller — leader and joiners alike — selects on completion vs its own context: each request's wait is bounded by its own deadline while the load still completes once into the cache for the callers that stay. `Manager.lookup` replaces `resilience/singleflight` (whose waiters ignore ctx by design) with a local `lookupFlight`; the cacheGen write-back race guard is unchanged. Panics in detached fills are converted to errors every waiter observes, since there is no caller left to re-panic into.

**P1 — `WithScope(nil)` silently ran unscoped.** Now a `NewManager` construction error: a caller who asked for scoping must never run without it (fail-closed tenancy contract).

**P1 — a resolver returning `(nil, nil)` passed creation, then panicked on the first hit.** Rejected at `Create`'s ref precheck, and `Manager.Handler` answers a logged 500 if one still reaches serve time (e.g. a row admitted via `SkipRefCheck`) — a consumer bug reads as an internal error, not a dead link or a crash.

**P2 — fixed-target links had no synchronous decorator seam.** New `WithDecorators(ds ...Decorator)` wraps every link's Decider, Target- and Ref-backed alike, composed with `Chain`. Ref-specific chains in `Cache.Resolver` still work and run inside the manager chain.

**P2 — `WithBaseURL` accepted query/fragment URLs that `ShortURL` rendered into unusable text** (`...?utm=x/promo`). Rejected at construction; the check runs on the raw string so a bare trailing `?` or `#` (which parses to empty `RawQuery`/`Fragment`) can't slip through.

**P2 — the ref compile cache had unbounded lifetime and cardinality.** Entries now expire after a TTL (default 5m; must be positive, mirroring `WithCache`'s bounded-staleness rule) and an amortized once-per-TTL sweep prunes refs that stop being clicked — bounding both the staleness of a missed `Invalidate` (or another node's save) and the resident set. **API change:** `NewCache(load, ...CacheOption)` with `WithRefTTL` and `WithCompileOptions` replaces the old variadic `Option` list. TTL expiry shares the compile options' clock, so it is testable via `WithClock`.

Reviewed and deliberately not changed: the global `cacheGen` counter (false-positive self-evictions cost one extra miss, never correctness; per-code generations would add a second keyed map), per-hit Target recompiles (documented decision backed by the ~µs compile benchmark), and the unscoped Postgres `List` sort (needs its own index; store-level trade-off).

### Benchmarks (Apple M3 Max, before → after)

|---|---|---|
| HandlerTarget | 3000 ns/op, 44 allocs | 3064 ns/op, 44 allocs |
| ResolveDecideTarget | 830 ns/op, 15 allocs | 860 ns/op, 15 allocs |
| HandlerRef | 1849 ns/op, 30 allocs | 1984 ns/op, 30 allocs |

Allocations unchanged; the HandlerRef delta is the TTL-expiry check on the ref cache hit path — the cost of bounded entry lifetime.

## Checklist

- [x] `just lint` passes
- [x] `just test` passes
- [x] New code has tests

### DIFF
--- a/auth/jwt/jwks.go
+++ b/auth/jwt/jwks.go
@@ -123,39 +123,34 @@ type jwksCache struct {
 	fetched bool
 }
 
-// needFetch reports whether resolving kid requires a network fetch given the
-// cache state at now: a cold cache, a set past its refresh TTL, or an unknown
-// kid past the refetch cooldown.
-func (c *jwksCache) needFetch(kid string, now time.Time) bool {
+// snapshot reads kid's resolution state under one lock acquisition: the key
+// (if present), whether the set was ever fetched, and whether resolving kid
+// requires a network fetch given now — a cold cache, a set past its refresh
+// TTL, or an unknown kid past the refetch cooldown.
+func (c *jwksCache) snapshot(kid string, now time.Time) (k Key, ok, fetched, need bool) {
 	c.mu.RLock()
-	_, ok := c.keys[kid]
+	k, ok = c.keys[kid]
 	fetched, fetchedAt := c.fetched, c.fetchedAt
 	c.mu.RUnlock()
 
-	if !fetched || now.Sub(fetchedAt) >= c.refresh {
-		return true
-	}
-	return !ok && now.Sub(fetchedAt) >= c.cooldown
+	need = !fetched || now.Sub(fetchedAt) >= c.refresh ||
+		(!ok && now.Sub(fetchedAt) >= c.cooldown)
+	return k, ok, fetched, need
 }
 
 func (c *jwksCache) get(ctx context.Context, kid string) (Key, error) {
-	c.mu.RLock()
-	k, ok := c.keys[kid]
-	fetched := c.fetched
-	c.mu.RUnlock()
-
-	need := c.needFetch(kid, c.clk.Now())
+	k, ok, fetched, need := c.snapshot(kid, c.clk.Now())
 	if ok && !need {
 		return k, nil
 	}
 	if need {
 		_, _, err := c.group.Do(ctx, "fetch", func(ctx context.Context) (struct{}, error) {
-			// Re-check under the flight: between a caller's staleness check and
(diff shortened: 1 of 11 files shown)


dmitrymomot:
### Review feedback and CI addressed

- Fixed the doc.go wrapping nit: rewrapped the Hooks paragraph to the file's ~78-column comment width (f878226).
- Fixed the CI `test` job failure — `TestJWKSConcurrentVerifySingleFetch` in auth/jwt (a package this PR doesn't otherwise touch) flaked because of a real TOCTOU: a caller could snapshot a stale cache, then become a fresh singleflight leader right after the previous fetch completed and deregistered, repeating the network fetch. The leader now re-checks staleness inside the flight and skips the network when a completed fetch already satisfied it (83066b5). Verified with `GOMAXPROCS=2 go test -race -run TestJWKS -count=500`.

@claude ready for re-review.

dmitrymomot:
### Non-blocking note addressed

Folded the JWKS staleness check into a single `snapshot(kid, now)` helper (1dc9d8d): `get()` is back to one `RLock` acquisition per call, and the fast-path decision now reads key presence, `fetched`, and staleness from one consistent view instead of two independent reads. Verified with `-race` (full package) plus `GOMAXPROCS=2 -run TestJWKS -count=200`; `just lint` clean.

@claude ready for re-review.

PR #77, opened 2026-07-18
feat(examples): redirector — full-stack smartlink demo (Postgres, geo routing, fingerprint dedup)

dmitrymomot:
## Description

Adds `examples/redirector`, a fully working short-link service demonstrating how the web/data/ops packages compose into a real app:

- **One port, two hosts** via `hostrouter` over lvh.me (resolves to 127.0.0.1): `go.lvh.me:8080/{code}` serves redirects, `app.lvh.me:8080` the dashboard.
- **Postgres-backed everywhere**: links through `smartlink/pgstore`; offers and clicks in the example's own tables. `migration.Group` applies the pgstore set and the app set under separate goose version tables against the bundled `postgres:18` docker-compose service (note: the 18 image moved its volume mount to `/var/lib/postgresql`).
- **Geo-aware routing**: offers (name + default URL + ordered geo rules) hydrate into `smartlink.Spec` through `smartlink.NewCache`, with `Cache.Invalidate` on the save path. Country resolves through a `geoip.Chain`: `?geo=XX` / `X-Geo-Country` overrides for local testing, `CF-IPCountry` behind Cloudflare, and an optional real IP lookup via `geoip/mmdb` when `GEOIP_MMDB` points at a MaxMind City database.
- **Click tracking with fingerprint dedup**: the `web/fingerprint` hash (headers + client hints + IP) doubles as the Visit `StickyKey`, so split bucketing and duplicate detection share one identity. `WithOnHit` does a non-blocking send into a bounded channel; a single-writer supervisor service inserts clicks with `is_duplicate` computed inside the INSERT (race-free by the single-writer construction). Dashboard stats (totals / uniques / dups, by-country, by-rule, recent) are SQL aggregates, so they survive restarts.
- **No seeded data**: offers and links are created through the dashboard forms (std `html/template` + `render.HTML`), with validation errors re-rendered as 422s.
- Own `justfile` (`just run` = compose up --wait + go run) and `docker-compose.yml`.

## Verified

Everything exercised live, not just compiled: forms create offers/links (bad country codes, ref+target both set, reserved/duplicate vanity codes all rejected); `?geo=de|us` routes to the matching rule and forwards sub-IDs + stamped campaign metadata (`ParamsFill`); repeat clicks from the same fingerprint count as duplicates while a different UA/IP counts unique; data survives an app restart; the mmdb path verified against the repo's MaxMind test fixture (`216.160.83.56` → US → matched the US/CA rule). A quick `ab` run sustained ~22k redirects/s (p99 11ms) with the bounded click sink dropping (and logging) overflow — the documented WithOnHit failure mode.

Reviewer notes: this is an example `main`, so there are no tests by design (repo packages it wires are tested); the example is excluded from the benchmarks policy for the same reason. The fingerprint secret is a fixed demo value, called out in a comment.

## Checklist

- [x] `just lint` passes
- [x] `just test` passes
- [ ] New code has tests — n/a: runnable example `main`, exercised end-to-end manually (see Verified)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -13,6 +13,28 @@ gracefully on `Ctrl+C` (SIGINT) or SIGTERM.
 go run ./examples/worker
 ```
 
+## redirector
+
+A fully working, Postgres-backed short-link service built from [`smartlink`](../web/smartlink) + [`smartlink/pgstore`](../web/smartlink/pgstore), [`fingerprint`](../web/fingerprint), [`geoip`](../web/geoip), [`useragent`](../web/useragent), [`hostrouter`](../web/hostrouter), [`data/postgres`](../data/postgres) + [`data/migration`](../data/migration), and [`render`](../web/render) (std `html/template`): geo-aware smart routing, click tracking with fingerprint-based duplicate detection, and a minimal dashboard with per-link stats plus create-offer and create-ref-link pages. One port, two hosts via `lvh.me` (resolves to 127.0.0.1): `go.lvh.me:8080` serves redirects, `app.lvh.me:8080` the dashboard.
+
+Links, offers, and clicks all live in Postgres (docker-compose ships a `postgres:18` service; both migration sets apply automatically on startup). Nothing is seeded — create an offer and a link through the dashboard forms, then click around:
+
+```sh
+cd examples/redirector
+just run          # docker compose up --wait + go run .
+# open http://app.lvh.me:8080/ — create an offer, then a link, then:
+curl -i "http://go.lvh.me:8080/<code>?geo=de"   # geo rule
+curl -i "http://go.lvh.me:8080/<code>?sub=abc"  # default target, sub-ID forwarded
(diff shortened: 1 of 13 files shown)


dmitrymomot:
### Review feedback addressed (e4e55db)

- **CSRF on dashboard forms** — wired `web/csrf` (double-submit over a `web/cookie` codec + `crypto/keyset`) around the dashboard host; both forms now carry a hidden `csrf_token` field. Verified live: POST without a token → 403, with cookie+token → 303 and the record lands; the redirect host is unwrapped and unaffected. The plain-HTTP demo drops `Secure` and the default `__Host-` cookie name (browsers reject `__Host-` over HTTP) — both called out in a comment as demo-only deviations.
- **`groupCount` SQL concat shape** — removed the interpolated fragment; the two dimensions are now complete constant queries (`clicksByCountrySQL`, `clicksByRuleSQL`) passed whole, so no SQL is built by concatenation anywhere.

@claude ready for re-review.

PR #79, opened 2026-07-18
feat(web/tenant): transport-layer tenant resolution package

dmitrymomot:
## Summary

Rebuild of the removed `data/tenant` as a **pure transport-layer** package at `web/tenant` — it derives the tenant from the request, validates it through consumer-owned seams, and returns the canonical tenant ID. No SQL query builders: how the ID scopes storage is each consumer package's concern (wire `tenant.Scope` into their `WithScope` options).

### Shape

- **Sources** (`Source func(*http.Request) (string, error)`): `Header`, `Cookie`, `Query`, `PathPrefix`, `Subdomain(base, lookup)`, `Domain(lookup)`, `Context()` passthrough, and `Map` for alias→ID translation on any other source. Subdomain labels / custom domains / slugs are aliases, never IDs — they translate through the storage-agnostic `SubdomainLookup`/`DomainLookup` seams (`StaticDomains`/`StaticSubdomains` in-memory impls for tests/dev).
- **Constructor**: `New(...Option) *Resolver` — `WithSources` (precedence chain, first non-empty wins), `WithValidator` (the soft-deleted/disabled/suspended status seam: `ErrTenantNotFound`/`ErrTenantInactive`, always fail-closed), `WithErrorHandler` (replaceable response, default 404 for rejections / 500 for infrastructure), `WithLogger` (default no-op).
- **Middleware**: stamps the resolved ID on the context; unresolved requests pass through untenanted (add `Require` where tenancy is mandatory); resolved-but-invalid tenants are rejected, never passed through. "Nothing resolved" is distinguished from errors **structurally**, so a consumer source/validator returning `ErrNoTenant` cannot forge the untenanted passthrough — it fails closed as 500.
- **Carrier**: `NewContext`/`FromContext`/`Scope` over `core/ctxkey`; `Scope` is the fail-closed hook matching every forge package's `WithScope(func(ctx) (string, error))`.

The source funcs and `normalizeHost` are a near-verbatim port of the reviewed data/tenant resolvers (PR #67); new code is the engine (`Resolver`/`New`/`Resolve`/`Middleware`), options, validator seam, and the `Query` source.

### Benchmarks (Apple M3 Max)

| bench | ns/op | allocs/op |
|---|---|---|
| Subdomain (hit) | 34.6 | 0 |
| Subdomain (miss) | 22.5 | 0 |
| Header | 19.5 | 0 |
| PathPrefix | 2.7 | 0 |
| Domain (static) | 39.6 | 0 |
| Resolve (2-source chain + validator) | 100 | 0 |
| Middleware end-to-end | 166 | 3 (ctx stamp + request clone, intrinsic) |

Numbers match the pre-removal package (e.g. old BenchmarkMiddleware was the same 3-alloc shape); no post-bench optimization was warranted — every parse/derive path is already zero-alloc.

### Testing

- 100% statement coverage, black-box (`tenant_test`), `-race` clean.
- Fuzz: `FuzzSubdomain` (label extraction properties, echo lookup) and `FuzzPathPrefix` — 10s sessions clean.
- Local agent review applied: structural fail-closed passthrough (above), security note on client-controlled sources overriding upstream identity (order `Context()` first), client-controlled caveat on `PathPrefix`, Debug-vs-Error log-level tests, `WithLogger(nil)` test.
- `just lint` (vet, golangci-lint, nilaway, betteralign, modernize, integration-tag tier) clean.

### DIFF
--- a/web/tenant/bench_test.go
+++ b/web/tenant/bench_test.go
@@ -1,0 +1,106 @@
+package tenant_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/dmitrymomot/forge/web/tenant"
+)
+
+func BenchmarkSubdomain(b *testing.B) {
+	derive := tenant.Subdomain("app.example.com", tenant.StaticSubdomains(map[string]string{"acme": "t_01acme"}))
+	r := newRequest("acme.app.example.com", "/")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := derive(r); id == "" {
+			b.Fatal("expected resolution")
+		}
+	}
+}
+
+func BenchmarkSubdomainMiss(b *testing.B) {
+	derive := tenant.Subdomain("app.example.com", tenant.StaticSubdomains(map[string]string{"acme": "t_01acme"}))
+	r := newRequest("other.example.org", "/")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := derive(r); id != "" {
+			b.Fatal("unexpected resolution")
+		}
+	}
+}
+
+func BenchmarkHeader(b *testing.B) {
+	derive := tenant.Header("X-Tenant-ID")
+	r := newRequest("example.com", "/")
+	r.Header.Set("X-Tenant-ID", "acme")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := derive(r); id == "" {
+			b.Fatal("expected resolution")
+		}
+	}
+}
+
+func BenchmarkPathPrefix(b *testing.B) {
+	derive := tenant.PathPrefix("/t")
+	r := newRequest("example.com", "/t/acme/dashboard")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := derive(r); id == "" {
+			b.Fatal("expected resolution")
+		}
+	}
+}
+
+func BenchmarkDomainStatic(b *testing.B) {
+	derive := tenant.Domain(tenant.StaticDomains(map[string]string{"shop.acme.com": "acme"}))
+	r := newRequest("shop.acme.com", "/")
+	b.ReportAllocs()
+	for b.Loop() {
+		if id, _ := derive(r); id == "" {
+			b.Fatal("expected resolution")
+		}
+	}
+}
+
+func BenchmarkResolve(b *testing.B) {
+	rv := tenant.New(
+		tenant.WithSources(
+			tenant.Domain(tenant.StaticDomains(map[string]string{"shop.acme.com": "t_01acme"})),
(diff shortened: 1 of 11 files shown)


dmitrymomot:
Addressed both review nits in 44575d1: `resolve` now uses plain (unnamed) results so the shadowing trap is gone, and `BenchmarkQuery` keeps the `url.Values` parse cost visible (131.6 ns/op, 416 B/op, 3 allocs/op — left unoptimized per the perf rules since it's not a proven-hot path). Both threads resolved.

dmitrymomot:
Reworked in eafa660 per discussion: the package now mirrors `auth/guard`'s two-phase split — pure extraction, then one resolution seam that looks up **and** validates in a single call.

- `Source` is now `func(*http.Request) (Identifier, bool)` — pure, no I/O, no error. Each shipped source tags its value with a `Kind` (`KindID`/`KindDomain`/`KindSubdomain`/`KindPath`; open string type, custom sources may invent kinds) so one consumer `Lookup` can interpret every namespace.
- `New(lookup, opts...)` takes the single required `Lookup` seam positionally (like `guard.New(verifier)`): `LookupTenant(ctx, Identifier) (string, error)` answers "which tenant" and "may it serve" in one call — typically one query with `AND deleted_at IS NULL AND suspended_at IS NULL`. The double lookup+validate round-trip is gone.
- Deleted: `Validator`, `ValidatorFunc`, `Map`, `DomainLookup`, `SubdomainLookup`, `StaticDomains`, `StaticSubdomains` (−165 net LOC).
- Semantics: `ErrTenantNotFound` continues the chain (a host is ambient — the marketing domain must fall through to untenanted passthrough + `Require`); `ErrTenantInactive` and infra errors fail closed; a lookup returning `("", nil)` is treated as a consumer bug and fails closed (guard's empty-Subject precedent).
- `Query` now scans `RawQuery` directly (guard.Query precedent): **131.6→77.0 ns/op, 3→0 allocs**. `Resolve` chain 100→62.7 ns/op, `Middleware` 166→136 ns/op. New `FuzzQuery` differential-checks the scan against `url.ParseQuery`; `FuzzPathPrefix` caught `/t//x` extracting an empty segment as present — fixed to not-present.
- 100% coverage retained; `-race`, all three fuzz targets, and `just lint` clean.

dmitrymomot:
Fixed both minor observations in 8a7b79a: `Query` now documents the 10,000-parameter-cap divergence (same caveat as `guard.Query` — extraction is not a bypass since the Lookup gates resolution). For `Context()`: the round-trip through the Lookup is intentional — it keeps a mid-session suspension effective on the next request — but the old doc claimed a "fast-path a trusted upstream" that `KindID` can't express (context- and header-derived IDs are indistinguishable to the Lookup), so the doc now states the real contract and points consumers who want an exempting fast-path at a custom source with its own `Kind`.

PR #80, opened 2026-07-18
feat(async/queue): add MongoDB broker driver

dmitrymomot:
## Description

Adds `async/queue/mongo` (package `mongoqueue`): a MongoDB `queue.Broker` + `queue.TxPusher` driver for the shipped async/queue engine, passing the full `brokertest` conformance suite under `-race` (85% coverage). Also bumps the shared `testkit/mongotest` container from the stale `mongo:7` pin to the current stable `mongo:8.3` (both mongo-backed integration suites pass against it).

### Design

- **Single collection, `state` discriminator (live/dead).** Kill, Requeue, and Purge are single atomic document updates — no cross-collection move can lose a job mid-crash, and everything except multi-job Push works on a standalone server.
- **One always-set indexed `visible_at` watermark** (run_at on push → lease deadline while claimed → retry time after nack). The claim scan is a single tight `{queue, visible_at, _id}` partial-index range; in-flight jobs never get fetched-and-filtered per poll, unlike the naive `$or` over an optional lease field. Watermarks are ceiled to the next millisecond so BSON's ms floor can't make a nacked job claimable before `retryAt` or expire a lease early.
- **Claiming.** Batch path: shortlist of candidate ids in visibility order → guarded `updateMany` stamping lease + fencing token + attempt atomically per document → fetch of the won documents, with bounded refill rounds when a concurrent claimer steals candidates (the engine reads a short `Claim` return as "queue drained"). Single-job claims are one atomic `findAndModify` — wall-time ties the batch path but 44% fewer allocations, a third of the server commands, and no claimed-but-unfetched crash window.
- **Transactional enqueue.** Multi-job Push wraps in its own session transaction to honor the all-or-nothing contract (replica set required); `PushTx` takes a caller-owned `*mongo.Session`, and Push joins an ambient transaction-bound context (`data/mongo.WithTransaction`) automatically. Both guard with `Session.TransactionRunning()` so a session bound for causal consistency only can never silently degrade a batch to a non-atomic ordered insert.
- **Primary-pinned reads.** The collection handle overrides the client's read preference: a queue must read its own acknowledged writes, and a `secondaryPreferred` client would otherwise lose claimed jobs between the guarded update and the fetch (leases burned with zero executions).
- Stats counts are bounded at 10k per queue with `Capped` flags (pgqueue semantics); retention sweeps drain in bounded 5000-doc rounds so a huge backlog costs many short commands instead of one unkillable delete.

### Review

An 8-angle adversarial review ran before pushing; all 10 verified findings are fixed in the second commit — the most notable: session-without-running-transaction silently dropping all-or-nothing, `secondaryPreferred` losing claimed jobs on the post-update fetch, BSON ms-floor breaching the `retryAt` contract, lease stamped from a pre-round-trip clock, and `ModifiedCount` used where `MatchedCount` is the ownership signal.

### Benchmarks (Apple M3 Max, mongo container, `-benchtime=2s`)

| benchmark | ns/op | B/op | allocs/op |
|---|---|---|---|
| PushClaimAck | ~1.85ms | 31,690 | 352 |
| PushBatch (100 jobs, txn) | ~2.5ms | 731,259 | 4,046 |
| ClaimBatch (claim 20 + acks, 20k backlog) | ~10.7ms | 387,610 | 3,251 |
| ListDead (50 of 5k) | ~0.4ms | 326,218 | 2,938 |

Post-benchmark optimization pass: the `findAndModify` n==1 fast path (kept on measured allocation/command wins after an honest same-environment A/B showed wall-time parity — the initial cross-container comparison had claimed 19%), hoisted constant BSON fragments, `SetBatchSize` on multi-doc finds to kill a `getMore` round trip past 101 docs, and client-side batch ordering to spare the server an uncovered sort stage.

### Notes for review

- `docs/design.md` driver enumerations and `docs/packages.md`'s shipped-driver parenthetical now include mongo.
- Collection name validation runs in the Docker-free unit tier (driver v2 `Connect` does no I/O); everything else is `//go:build integration` via `testkit/mongotest` (single-node replica set, so multi-document transactions work).
- Client-clock visibility semantics (like `queue.MemoryBroker`) are documented in `doc.go`: fencing keeps delivery at-least-once regardless of worker clock skew.

## Checklist

- [x] `just lint` passes
- [x] `just test` passes
- [x] New code has tests

### DIFF
--- a/async/queue/mongo/bench_test.go
+++ b/async/queue/mongo/bench_test.go
@@ -61,15 +61,21 @@ func BenchmarkMongoClaimBatch(b *testing.B) {
 	broker := newBroker(b)
 	ctx := context.Background()
 
-	// Seed enough backlog that Claim never runs dry inside the measured loop.
+	// seed keeps a backlog ahead of the measured loop; when it runs dry
+	// (long -benchtime or -count), it is refilled outside the timer so any
+	// iteration count measures pure claim/ack work.
 	const seed = 20000
-	jobs := make([]queue.Job, seed)
-	for i := range jobs {
-		jobs[i] = benchJob("default")
+	refill := func() {
+		b.Helper()
+		jobs := make([]queue.Job, seed)
+		for i := range jobs {
+			jobs[i] = benchJob("default")
+		}
+		if err := broker.Push(ctx, jobs...); err != nil {
+			b.Fatal(err)
+		}
 	}
-	if err := broker.Push(ctx, jobs...); err != nil {
-		b.Fatal(err)
-	}
+	refill()
 
 	b.ReportAllocs()
 	for b.Loop() {
@@ -78,7 +84,10 @@ func BenchmarkMongoClaimBatch(b *testing.B) {
 			b.Fatal(err)
 		}
 		if len(got) == 0 {
-			b.Fatal("backlog ran dry mid-benchmark")
+			b.StopTimer()
+			refill()
+			b.StartTimer()
+			continue
 		}
 		for _, j := range got {
 			if err := broker.Ack(ctx, j.ID, j.Token); err != nil {
--- a/async/queue/mongo/bench_test.go
+++ b/async/queue/mongo/bench_test.go
@@ -1,0 +1,125 @@
+//go:build integration
+
+package mongoqueue_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/dmitrymomot/forge/async/queue"
+	"github.com/dmitrymomot/forge/core/id"
+)
+
+// Past-biased RunAt throughout so every pushed job is immediately claimable.
+func benchJob(q string) queue.Job {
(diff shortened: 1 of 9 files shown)


dmitrymomot:
### Review feedback addressed

- Fixed the design.md contradiction (336de4f): the PushTx policy line now states the split by what actually gates it — real multi-document ACID transactions — instead of SQL vs non-SQL. Kept `mongoqueue`'s native `TxPusher`: MongoDB multi-document transactions are the real thing, and the implementation guards against the degraded case (`Session.TransactionRunning()`), so dropping it in favor of async/outbox would remove a working capability to satisfy stale wording. Updated the matching sentence in packages.md and redisqueue's doc.go pointer (`queue/mongo` is now a transactional-enqueue alternative) in the same commit.

@claude ready for re-review.

PR #82, opened 2026-07-20
feat(comms/postback): tracker-style S2S postback sender

dmitrymomot:
## Summary

New `comms/postback` package — the affiliate/ad-network conversion ping: unsigned GET (or POST) to a partner-configured URL template whose `{macro}` placeholders carry click ID, payout, status, and sub-IDs.

- **`Vocabulary`** — caller-registered macro names (consumer data; forge ships none). Names validated fail-closed (`ErrInvalidMacro`); the zero value rejects every macro.
- **`Destination`** — URL template parsed against a vocabulary at registration, fail-closed: unknown macro (`ErrUnknownMacro`), unbalanced braces, non-http(s)/relative URL, fragment, or a macro in the scheme/authority (`ErrInvalidTemplate`) are construction errors — never an empty substitution at fire time. Method GET default, POST via `WithMethod` (`ErrInvalidMethod` otherwise).
- **`Sender.Send`** — renders the per-event value map with positional URL escaping (query `QueryEscape`, path `PathEscape` + `:` + dot-segment values `%2E`-encoded so a hostile value can never alter the URL structure validated at registration), fires through one reused `*http.Client`, and reports outcome by status class: **nil for 2xx, `ErrServerStatus` for 5xx (transient — requeue), `ErrClientStatus` otherwise (permanent — dead-letter)**. `Result{URL, StatusCode}` feeds the audit log.
- Registered-but-absent values render empty by design — sub-IDs are genuinely sparse and trackers accept empty parameters.
- Default client is `httpclient.New` with a 10s timeout (retries transient failures on GET, not POST); `WithHTTPClient` is the seam for custom retry/breaker policy.

Per the catalog design: unsigned (trackers correlate by click ID; HMAC deliveries are `comms/webhook`), no dedup (stable event IDs ride as macros; receivers dedup), no durability (rides `async/queue`/`async/eventrouter` as a deliverer), no tenant seam (a `Destination` is a passed value). The macro vocabulary stays consumer data, so per-tracker format tables (pending from Alisa) don't block this package.

Catalog entry deleted from `docs/packages.md` per the shipped-package rule; `async/eventrouter` deps line updated.

## Review pass

An adversarial pre-PR review caught three real issues, all fixed with regression tests:
1. Zero-value `Sender` panicked on nil client — now fails closed with an error.
2. A macro in or before the scheme (`{click_id}https://…`, `http{click_id}://…`) passed registration fail-open — `escapeModeFor` now treats a prefix without `://` as scheme/authority position and rejects it.
3. A path-position value of `.`/`..` rendered a dot-segment that path-normalizing servers collapse — bare dot values are now `%2E`-encoded (percent-encoded dots are not dot-segments, RFC 3986 §3.3).

## Benchmarks (Apple M3 Max)

```
BenchmarkNewDestination-14   2560393    455.1 ns/op    960 B/op    8 allocs/op   (cold path, registration)
BenchmarkRender-14          10883851    112.0 ns/op    112 B/op    1 allocs/op   (hot path — the 1 alloc is the result string)
BenchmarkSend-14               32976   35317 ns/op    4922 B/op   57 allocs/op   (dominated by net/http over httptest)
```

No post-benchmark optimization applied: `Render` is already at the 1-alloc floor (`strings.Builder` grown once; `QueryEscape` returns its input unchanged for clean values), and `Send`'s allocations are all `net/http` internals.

## Testing

- `go test -race ./comms/postback/` — black-box tests: template parse/validation matrix, positional escaping (incl. structure-injection and traversal attempts), status-class mapping, POST method, transport failure, context cancellation, zero-value `Sender`/`Destination` fail-closed.
- `just lint` — clean (vet, golangci-lint, betteralign, nilaway).

### DIFF
--- a/comms/postback/bench_test.go
+++ b/comms/postback/bench_test.go
@@ -49,6 +49,23 @@ func BenchmarkRender(b *testing.B) {
 	}
 }
 
+func BenchmarkRenderPathMacro(b *testing.B) {
+	vocab := benchVocab(b)
+	dest, err := postback.NewDestination("https://tracker.example.com/pb/{click_id}/{status}?s1={sub1}", vocab)
+	if err != nil {
+		b.Fatal(err)
+	}
+	values := map[string]string{
+		"click_id": "8f14e45fceea167a",
+		"status":   "approved",
+		"sub1":     "campaign-7",
+	}
+	b.ReportAllocs()
+	for b.Loop() {
+		_ = dest.Render(values)
+	}
+}
+
 func BenchmarkSend(b *testing.B) {
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
--- a/comms/postback/bench_test.go
+++ b/comms/postback/bench_test.go
@@ -1,0 +1,70 @@
+package postback_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/dmitrymomot/forge/comms/postback"
+)
+
+func benchVocab(b *testing.B) postback.Vocabulary {
+	b.Helper()
+	v, err := postback.NewVocabulary("click_id", "payout", "status", "sub1", "sub2")
+	if err != nil {
+		b.Fatal(err)
+	}
+	return v
+}
+
+func BenchmarkNewDestination(b *testing.B) {
+	vocab := benchVocab(b)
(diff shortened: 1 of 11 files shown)


dmitrymomot:
Addressed the three nit-level items from the review in 948d41a: added a test that `WithHTTPClient(nil)` falls back to the default client, added `BenchmarkRenderPathMacro` so path-position escaping is exercised (94.94 ns/op, 80 B/op, 1 alloc/op — also at the 1-alloc floor), and made the `:` → `%3A` comment precise (it applies to every path-position macro; over-encoding outside the first segment is harmless).

PR #88, opened 2026-07-20
feat(async): add outbox package — transactional outbox over any queue.Broker

dmitrymomot:
## Summary

Implements `async/outbox` from the catalog: the transactional outbox — job intent rows committed inside the business DB transaction, plus a relay `supervisor.Service` that forwards committed rows into any `queue.Broker`. This is the bridge from a Postgres transaction to redis/nats/kafka delivery for brokers without native `queue.TxPusher`.

## Design

- **`outbox.Store`** — the storage seam: `Add` rides the caller's transaction (driver-specific `tx`, pgoutbox asserts `pgx.Tx`); `Claim` is lease-based and strictly pull; `Delete`/`Fail` finish or reschedule rows. Claims are deliberately **unfenced** (no claim token, unlike `queue.Broker`): the post-claim window is one push + one delete, and every lost-lease interleaving degrades to a duplicate push, which the at-least-once contract downstream already absorbs.
- **`outbox.Wrap(store, broker)`** — a `queue.Broker` decorator that adds `queue.TxPusher` backed by the store, so `queue.PushTx` and `eventbus.PublishTx` work unchanged; plain `Push` and all other ops delegate untouched. It also forwards `queue.Maintainer` by delegation so wrapping doesn't silently disable the inner broker's housekeeping when the worker probes for the capability.
- **`outbox.NewRelay(store, broker)`** — a `supervisor.Service`: claim batch under lease → `Push` → delete. Batch push failure falls back to per-row push so one poison row can't wedge healthy rows. Failed rows are rescheduled with capped exponential backoff (5s → 5m, 20% jitter) and **never dropped** — outbox delivery is guaranteed, and a broker outage resolves by draining the backlog. Claim errors and zero-progress cycles widen the poll interval instead of hammering a down store/broker (mirrors `queue.Service`).
- **`MemoryStore`** — in-process test double (tx ignored, documented as non-transactional); **`async/outbox/postgres`** (`pgoutbox`) — the real store: unnest batch insert on the caller's `pgx.Tx`, `FOR UPDATE SKIP LOCKED` claim (any number of relay instances, no coordination), goose migration exposed as `pgoutbox.Migrations`.
- **Delayed jobs**: rows forward immediately regardless of `RunAt` — the queue engine owns delay semantics.
- **Tenancy**: no outbox-level seam needed — the scope is captured fail-closed by the producing `queue.Client`/`eventbus`, and outbox stores/forwards the envelope verbatim (asserted in the pg roundtrip test). Single-tenant use pays zero ceremony.

## Correctness notes

- `available_at` is stamped by the **database clock** (`now()`) at insert, not the client's `CreatedAt` — a client clock running ahead of Postgres would otherwise delay every row by the skew (the clock-skew class of flake we hit with the mongo driver).
- Uncommitted rows are invisible to `Claim`; rollback leaves nothing behind (tested).
- Concurrent claim overlap test: 20 rows, 4 workers, every row claimed exactly once.

## Benchmarks (Apple M3 Max, postgres:18-alpine via testcontainers)

| bench | ns/op | B/op | allocs/op |
|---|---|---|---|
| MemoryStore_Add (100 rows) | ~13,000 | 22,400 | 100 |
| MemoryStore_ClaimDelete (100 rows) | ~34,000 | 43,777 | 102 |
| Broker_PushTx (100 rows, wrapper overhead) | ~5,600 | 22,400 | 100 |
| PgOutbox_Add100 (batch insert in tx) | ~1.09ms | 161,600 | 1,574 |
| PgOutbox_ClaimDelete100 | ~1.27ms | 148,600 | 1,154 |

Post-benchmark pass: all pg ops are already single-statement / single-round-trip (unnest insert, one CTE claim, `ANY(uuid[])` delete), and the numbers are round-trip dominated — ~10.9µs/row insert vs pgqueue's measured profile for the same shape. No perf-motivated complexity added; the only change out of the pass was the DB-clock `available_at` stamp, which is correctness, not speed. `CopyFrom` switching (pgqueue's ≥2000-row optimization) was deliberately skipped: outbox batches are business-transaction-sized, so the crossover is unreachable in practice.

## Tests

- Unit (Docker-free): memory store contract, wrapper delegation + `PushTx`/`Maintain`, relay forward/drain, broker-down-then-recover, poison-row isolation, cancel; runnable `Example` wiring the full PushTx → relay → worker pipeline.
- Integration (`just test-integration`): pg store contract incl. commit/rollback visibility, lease expiry reclaim, fail rescheduling, custom table, relay end-to-end over pg, concurrent claim no-overlap. All passing locally with `-race`.


dmitrymomot:
### Review feedback addressed

- **Claim index mismatch** ([migration](https://github.com/dmitrymomot/forge/blob/dm/async-outbox-package-2d8a77/async/outbox/postgres/migrations/20260720120000_outbox_jobs.sql)): rather than bolting `available_at` onto the old index, the claim's pick order now follows pgqueue's own pattern — `picked` selects by `(available_at, id)` with a matching `outbox_jobs_claim_idx (available_at, id)`, making selection a pure bounded index range scan with no sort even when a retry backlog parks many rows with future `available_at` (an `(available_at, created_at, id)` index alone would still have forced a sort of every due row under `ORDER BY created_at`). The returned batch (≤ n rows) is re-sorted into `(created_at, id)`, so the Store contract's observable ordering is unchanged; contract wording and MemoryStore pick order updated to match.
- **Clock/backoff coverage** (non-blocking note): added `TestRelay_BackoffUsesInjectedClock` — mock clock frozen ⇒ a failed row stays parked across many poll cycles, advance past `retryAt` ⇒ it retries — pinning `retryAt = clk.Now() + bo.Next(attempts)` directly, plus a `WithLogger` test asserting relay log output.

Unit + integration suites pass with `-race`; `just lint` clean.

@claude ready for re-review.

dmitrymomot:
CI round two's `integration` failure was an unrelated pre-existing flake: `TestDoDetachedExecutionSurvivesCallerCancel` in `resilience/singleflight` races when an instant fn and a pre-cancelled ctx make both arms of `DoDetached`'s select ready — the runtime picks randomly, so the `context.Canceled` assertion was a coin flip. Reproduced locally with `GOMAXPROCS=2 -race` (the CI runner shape) within 500 runs; the product behavior is fine (returning an already-available result to a cancelled caller is benign), so the fix holds the flight open until the caller returns, making the assertion deterministic while still verifying the detached execution completes uncancelled. 500 repro-condition runs now pass.

dmitrymomot:
### Remaining minor review notes addressed

Every finding from all three review rounds is now fixed, including the explicitly non-blocking ones:

- **`MemoryStore.Claim` sort-all** (round 1, minor): now bounded selection — only the n most-overdue entries are kept in a sorted buffer, with O(1) rejection once the buffer is full. Measured (new `BenchmarkMemoryStore_ClaimFromLargeBacklog`, claim 100 of 10k): ~2.1ms / 377KB / 13 allocs → ~1.1ms / 21KB / 2 allocs per claim.
- **Divergent pick-vs-return order coverage** (round 2, very minor): added `TestMemoryStore_PickByOverdueReturnByCreated` and `TestPgOutbox_PickByOverdueReturnByCreated` — a retry backlog reverses `available_at` against `created_at`; both stores must pick the most-overdue rows first while returning the claimed batch ordered by `(created_at, id)`, and must leave the least-overdue row unpicked. The pg variant uses minute-scale offsets so a lagging container DB clock cannot flake it.

Previously fixed and verified by re-review: claim index/filter match, mock-clock backoff test, `WithLogger` test, singleflight flake. Unit + integration suites pass with `-race`; `just lint` clean.

@claude ready for re-review.

PR #89, opened 2026-07-20
feat(async/eventrouter): batched event egress over eventbus with per-destination isolation

dmitrymomot:
## Summary

`async/eventrouter` — event egress over `async/eventbus`: routes published events to external destinations (warehouses, partner endpoints, affiliate trackers) with per-destination isolation, batched delivery, and queue-grade retry.

- **One subscription per (event, destination)**: `Route(bus, evt, dest)` registers `dest.Name()` as an eventbus subscription, so each pair is its own durable queue — a slow or failing destination only delays itself.
- **Filter/remap as registered Go functions** (`WithFilter`, `WithRemap`) — no mapping DSL. A remap error (or unmarshalable remapped payload) is poison: `SkipRetry` dead-letter, no attempt burn.
- **Batched delivery with size+age flush**: events joining a destination accumulate into a batch (`WithBatchSize`, default 100; `WithBatchAge`, default 1s); the delivering jobs *block until the flush resolves*, so nothing is acked before the destination accepted it — at-least-once holds through crashes. `WithBatchSize(1)` disables batching (singleton fast path, right for sync buses and trackers).
- **Batch-level retry + poison isolation**: transient failure → every job in the batch retries on the queue's backoff (events re-batch on redelivery). Permanent failure (`Permanent(err)`) on a multi-event batch → each event re-delivered alone so the poison event dead-letters by itself; on a singleton → straight dead-letter.
- **Never dedups**: stable event IDs ride every delivery (per-event `id` field, `Idempotency-Key` header on single-event deliveries); receivers dedup — the Stripe contract. In-router suppression would trade duplicates for silent loss.
- **Reference deliverers** (engine only, no connector catalog): `NewHTTPDeliverer` (generic JSON-batch POST over `web/httpclient`), `NewWebhookDeliverer` (HMAC-signed via `comms/webhook`), `NewPostbackDeliverer` (tracker macro-URL pings via `comms/postback`, per-event with worst-outcome classification so isolation splits mixed batches precisely).
- **Tenancy**: optional fail-closed `WithScope` hook keys batches by tenant — batches never mix scopes, missing scope is a delivery error; single-tenant apps configure nothing. Deliverer panics are contained as retryable failures (a panicking deliverer must not crash a timer goroutine).

Catalog entry removed from `docs/packages.md` per the shipped-package rule.

## Benchmarks

Apple M3 Max, `-benchmem`:

| benchmark | ns/op | B/op | allocs/op |
|---|---|---|---|
| `BenchmarkPublishRouted` (sync publish → route → singleton deliver) | 941 | 704 | 16 |
| `BenchmarkPublishRoutedBatched` (8 concurrent publishers, size-8 flush) | 24,648 (per 8 events) | 4,943 | 107 |
| `BenchmarkBatchEncode` (100-event wire encoding) | 48,642 | 21,242 | 102 |

Baseline: raw `eventbus` sync publish with 1 subscription is 730 ns / 10 allocs, so the router adds ~210 ns / 6 allocs per event (payload re-marshal + Event construction + singleton fast path). Post-benchmark pass: no optimization applied — the overhead is noise against network-bound delivery, and the perf rules forbid unproven complexity.

## Testing

- Black-box tests: size flush, age flush, transient whole-batch retry, permanent singleton dead-letter, poison isolation (including mixed transient/permanent isolation verdicts), deliverer panic containment, delivery timeout, scope keying + fail-closed missing scope, filter/remap semantics, durable end-to-end over the memory broker (batching across jobs, retry keeps the stable event ID).
- Deliverer adapters tested against `httptest` servers: body/headers/signatures, status-class mapping, macro rendering, values-fn poison.
- `just lint` clean; `go test -race -count=3` clean.

### DIFF
--- a/async/eventrouter/bench_test.go
+++ b/async/eventrouter/bench_test.go
@@ -1,0 +1,81 @@
+package eventrouter_test
+
+import (
+	"context"
+	"encoding/json"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/dmitrymomot/forge/async/eventbus"
+	"github.com/dmitrymomot/forge/async/eventrouter"
+)
+
+type nopDeliverer struct{}
+
+func (nopDeliverer) Deliver(context.Context, []eventrouter.Event) error { return nil }
+
+// BenchmarkPublishRouted measures the full sync-bus publish → route → deliver
+// path with batching disabled: envelope marshal, handler dispatch, payload
+// re-marshal, and the destination's singleton fast path.
+func BenchmarkPublishRouted(b *testing.B) {
+	bus := eventbus.NewSync()
+	evt := eventbus.NewEvent[payload]("bench.routed")
+	dest := eventrouter.NewDestination("nop", nopDeliverer{}, eventrouter.WithBatchSize(1))
+	eventrouter.Route(bus, evt, dest)
+	ctx := context.Background()
+	p := payload{V: "bench"}
+
+	b.ReportAllocs()
+	for b.Loop() {
+		if err := eventbus.Publish(ctx, bus, evt, p); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+// BenchmarkPublishRoutedBatched measures the batching path under concurrent
+// publishers: joins coordinate through the destination mutex and flush by
+// size.
+func BenchmarkPublishRoutedBatched(b *testing.B) {
+	bus := eventbus.NewSync()
+	evt := eventbus.NewEvent[payload]("bench.batched")
+	dest := eventrouter.NewDestination("nop", nopDeliverer{},
+		eventrouter.WithBatchSize(8), eventrouter.WithBatchAge(time.Millisecond))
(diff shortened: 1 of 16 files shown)


dmitrymomot:
### Review feedback addressed

- **postback.go partial-success re-fire** — fixed in a482955, going further than the doc-only suggestion: `PostbackDeliverer.Deliver` now refuses any multi-event batch with a `Permanent` error *before firing anything*, which routes the batch through the router's existing poison-isolation pass — each event is then delivered alone with its own precise verdict. A batchmate's transient failure can no longer retry (and re-ping) events the tracker already accepted, at any batch size; `WithBatchSize(1)` just skips the refused batch attempt. Added a router-level test (`routed batch pings each event exactly once with precise verdicts`) asserting exactly-once pings and a per-event dead-letter on a mixed-outcome batch.

`just lint` clean, `go test -race -count=2` clean.

@claude ready for re-review.

PR #96, opened 2026-07-20
feat(finance/fxrate): exchange rates with recorded-rate conversions

dmitrymomot:
## Summary

New `finance/fxrate` package (first of the `finance/` namespace): exchange rates behind a `RateSource` seam with stored snapshots. `Convert` records the rate applied — provider, as-of time, value, scale, rounding mode — so audits answer "what rate at transaction time" and any stored `Conversion` or `Snapshot` recomputes byte-for-byte. Math is exact multiply-and-round via `core/decimal`; nothing routes through float64.

- **`Snapshot`** — immutable validated rate table: normalized codes, strictly positive rates, fail-closed construction (empty base/provider, zero as-of, no rates, duplicates, non-positive rates all rejected); the caller's map is copied. JSON round-trip goes back through `NewSnapshot`, so corrupt/tampered stored data fails closed. Inverse and cross rates derive by division at a fixed `RateScale = 12` half-even (package constant, not a knob) so derivations are deterministic forever; direct quotes keep the provider's scale.
- **`Converter`** — TTL-cached snapshot over the seam (`core/clock` for testable time), concurrent refreshes coalesced via `resilience/singleflight` with a post-flight freshness re-check. Fails closed: stale cache + failing source is an error, never silently-stale rates; a source returning the wrong base, a zero snapshot, or omitting a requested quote is rejected at fetch.
- **`StaticSource`** — fixed-rate source for tests and contractual rate cards (per-deal rates = one StaticSource per deal).
- **`frankfurter` subpackage** — thin JSON adapter over `web/httpclient` (ECB reference rates, no API key, GET retries via httpclient defaults): one GET, 1 MiB body cap, bare JSON numbers decoded straight into `core/decimal`. No provider SDKs, no live streaming.

Tenancy: rates are market data shared by all tenants — one Converter serves a multi-tenant app; per-tenant negotiated rates are separate Converter/StaticSource values chosen by the caller (passed-value pattern, per the calculator-package precedent).

Roadmap entry deleted from `docs/packages.md` per the shipped-package rule.

## Benchmarks (Apple M3 Max, Go 1.26)

```
BenchmarkSnapshotConvertDirect-14     2506455   427.4 ns/op   280 B/op   11 allocs/op
BenchmarkSnapshotConvertCross-14      1037280  1109 ns/op     704 B/op   27 allocs/op
BenchmarkSnapshotRateCross-14         2184255   508.3 ns/op   424 B/op   16 allocs/op
BenchmarkConverterConvertCached-14    3920846   304.0 ns/op   280 B/op   11 allocs/op
```

**Post-benchmark optimization pass:** alloc profile shows ~100% of allocations come from `core/decimal.Rescale`, which promotes to `math/big` on every scale reduction even when the coefficient fits int64 (`roundBig`/`pow10Big`); fxrate's own code contributes essentially zero. No in-package optimization is justified (readable-first rule); an int64 fast path for `decimal.Rescale` is flagged as a follow-up task, which would take the direct-convert path to ~0 allocs framework-wide.

## Testing

- Black-box tests: snapshot invariants, rate-derivation determinism (identity/direct/inverse/cross, half-even ties incl. negatives), JSON round-trip + fail-closed unmarshal, converter TTL/singleflight/fail-closed paths (`clock.Mock`), static source narrowing, frankfurter adapter over `httptest` (query shape, exact decimal decode, error paths).
- `go test -race ./finance/...` green; `just lint` 0 issues.

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -256,17 +256,6 @@ tests: `testkit/dbtest` (planned).
 
 ---
 
-**finance/fxrate**
-
-Exchange rates behind a `RateSource` seam with stored snapshots; `Convert`
-records the rate applied, so audits answer "what rate at transaction
-time". Math is multiply-and-round via `core/decimal`; providers are thin
-JSON adapters over httpclient — no provider SDKs, no live streaming.
-
-Deps: `core/decimal`, `web/httpclient`.
-
----
-
 **finance/tariff**
 
 Tiered/banded rate calculation over `core/money`/`core/decimal`:
--- a/finance/fxrate/bench_test.go
+++ b/finance/fxrate/bench_test.go
@@ -1,0 +1,84 @@
+package fxrate_test
+
+import (
+	"context"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/dmitrymomot/forge/core/decimal"
+	"github.com/dmitrymomot/forge/finance/fxrate"
+)
+
+func benchSnapshot(b *testing.B) fxrate.Snapshot {
+	b.Helper()
+	rates := map[string]decimal.Decimal{
+		"USD": decimal.MustParse("1.0850"),
+		"GBP": decimal.MustParse("0.8425"),
+		"JPY": decimal.MustParse("160.23"),
+		"CHF": decimal.MustParse("0.9782"),
+	}
+	// Pad to a realistic full-table size.
+	for i := range 26 {
+		rates["X"+strconv.Itoa(i)+"A"] = decimal.MustParse("1.2345")
+	}
+	s, err := fxrate.NewSnapshot("EUR", "bench", time.Date(2026, 7, 20, 0, 0, 0, 0, time.UTC), rates)
+	if err != nil {
+		b.Fatal(err)
+	}
+	return s
+}
+
+func BenchmarkSnapshotConvertDirect(b *testing.B) {
+	s := benchSnapshot(b)
+	amount := decimal.MustParse("99.90")
+	b.ReportAllocs()
+	for b.Loop() {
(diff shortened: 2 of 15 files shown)


dmitrymomot:
Addressed all three review findings:

- **`Do` → `DoDetached`** in `Converter.Snapshot`: followers joining an in-flight refresh are now released by their own context deadline while the shared fetch keeps running for callers that stay. Added `TestConverterFollowerWaitBoundedByOwnContext` (slow source, 50ms-deadline follower) and verified it fails with `Do` before the swap.
- **Provider trim** in `NewSnapshot`: whitespace-only provider now fails closed (`ErrInvalidSnapshot`), matching the doc.go claim; validation-table case added.
- **frankfurter per-quote trim**: was already fixed in an earlier commit (per-element `TrimSpace` + empty-element drop before building `symbols`), with the fixture test asserting normalized output.

`go test -race ./finance/...` and `just lint` green.

dmitrymomot:
Merged `origin/main` (conflict was the `docs/packages.md` roadmap section: ledger/tariff entries deleted on main vs fxrate deleted here — resolved to keep only `finance/formula`, which is still unshipped). Main's merge also pulled in the decimal Rescale fast path (PR #100), so the benchmark numbers from the PR description improve as predicted by its alloc profile:

```
BenchmarkSnapshotConvertDirect    427.4 ns/op  280 B/op  11 allocs/op  →  73.3 ns/op   0 B/op  0 allocs/op
BenchmarkConverterConvertCached   304.0 ns/op  280 B/op  11 allocs/op  →  127.9 ns/op  0 B/op  0 allocs/op
```

`go test -race ./finance/... ./core/decimal/` and `just lint` green on the merge result.

PR #97, opened 2026-07-20
feat(finance/invoice): invoice document model package

dmitrymomot:
## Summary

First package in the `finance/` namespace: the invoice document model — invariants, not rendering.

- **Document model**: `Invoice` is a plain, persistable struct; invariants are enforced at the package's operation boundary (the `fsm` philosophy of caller-owned state). A draft is freely mutable; `Issue` validates, freezes totals, draws the number, and moves draft → issued; every later correction is a new credit-note document via `NewCreditNote`, back-referencing the original through `Corrects` (corrections post forward, the rule shared with the planned ledger). `Verify` recomputes totals from lines to detect drift on loaded documents.
- **Numbering**: per-series `Sequence` over a small `SequenceStore` seam with two explicit, mechanically distinct modes — the requirement is jurisdictional. `Gapless` (default) draws only inside `Issue` (inside the caller's persistence tx when the store rides ctx) and refuses pre-drawn/pre-assigned numbers; `WithGaps` allows pre-drawing via `Sequence.Next` and burns numbers on failed issuance. `Issue` draws the number last, after every other invariant passed, so a validation failure never burns one (tested). In-memory store ships for tests/dev.
- **Totals**: line items → per-rate tax lines → totals over `core/money`, with `RoundPerLine` vs `RoundPerTotal` policies (banker's rounding, money's settlement default). Per-total mode rounds once at document level and allocates the subtotal back across lines penny-perfect via `money.Allocate`; `LineNets` always sum exactly to `Subtotal` (property-tested over 200 pseudo-random documents × both policies). Tax rates are caller-supplied data per line — never determined.
- **Lifecycle**: draft → issued → partially_paid/paid/overdue/void compiled once over `core/fsm`; voiding is guarded — a document with payments can only be corrected forward with a credit note (`fsm.ErrGuardDenied` wrapping `ErrHasPayments`). Payments match by external refs (ledger posting refs): identical replays are idempotent no-ops (at-least-once feeds safe), same ref with different amount is `ErrPaymentConflict`, overpayment records nothing.
- **Self-billing** is the `Direction` field (platform issues on the supplier's behalf — affiliate/agent payouts); a non-base-currency document records its `FXSnapshot` — recorded data, never conversion math.
- **Tenancy**: optional `WithScope` hook on `NewSequence`; fails closed with `ErrScope`, per-tenant counters use a length-prefixed storage key so (scope, series) pairs can never collide (tested), formatted numbers stay tenant-clean. Single-tenant use pays zero ceremony.

Removes the shipped `finance/invoice` entry from `docs/packages.md` per the roadmap rule.

## Anti-scope

No rendering (HTML is a `render` recipe, PDF consumer-side), no tax determination, no e-invoicing formats, no dunning, no subscription/pricing logic — the billing anti-scope stands.

## Benchmarks (Apple M3 Max)

Before/after for the one measured optimization — a single-line fast path in per-total allocation (single-line documents are the common SaaS case):

|---|---|---|
| Compute/per_total/1 | 663 ns/op, 944 B, 24 allocs | **312 ns/op, 544 B, 4 allocs** |

Full run after:

```
BenchmarkCompute/per_line/1-14      306.9 ns/op    544 B/op     4 allocs/op
BenchmarkCompute/per_total/1-14     311.6 ns/op    544 B/op     4 allocs/op
BenchmarkCompute/per_line/10-14     3517 ns/op    6371 B/op   111 allocs/op
BenchmarkCompute/per_total/10-14    4590 ns/op    7876 B/op   174 allocs/op
BenchmarkCompute/per_line/100-14   29445 ns/op   45891 B/op  1099 allocs/op
BenchmarkCompute/per_total/100-14  32978 ns/op   52687 B/op  1343 allocs/op
BenchmarkIssue-14                   3745 ns/op    7121 B/op   115 allocs/op
BenchmarkApplyPayment-14            186.4 ns/op    832 B/op     2 allocs/op
BenchmarkSequenceNext-14            128.2 ns/op     40 B/op     3 allocs/op
```

Remaining allocations are exact-decimal arithmetic inside `core/money`/`core/decimal`; issuance is a human-rate path, not a request hot path, so no further perf-motivated complexity.

## Testing

- `go test -race ./finance/...` — unit tests, black-box (`invoice_test`): totals known-value cases (the classic 3 × 33.33 @ 20% VAT divergence: 120.00 per-line vs 119.99 per-total), allocation penny-exactness, zero-minor-unit currency (JPY), sub-minor-unit fallback, full lifecycle matrix, numbering modes, credit notes, tenancy fail-closed + key-collision resistance, concurrent draw uniqueness, 200-document invariant fuzz.
- `just lint` — clean (golangci-lint, nilaway, betteralign, modernize, integration tags).

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -298,27 +298,6 @@ named Go functions with per-deal parameters as data.
 
 Deps: `core/decimal`.
 
----
-
-**finance/invoice**
-
-The invoice document model — invariants, not rendering: numbering via a
-per-series `Sequence` with two explicit modes (strict-gapless
-transactional counter vs monotonic-with-gaps — the requirement is
-jurisdictional); immutable once issued, corrections are credit notes
-back-referencing the original (the corrections-post-forward rule shared
-with `ledger`); line items → tax lines → totals in `money` with per-line
-vs per-total rounding policy via `Allocate`; draft → issued →
-paid/partially-paid/void/overdue over `fsm`, paid-matching by `ledger`
-posting refs; self-billing direction (platform issues on the supplier's
-behalf — affiliate/agent payouts); multi-currency with the `fxrate`
-snapshot recorded. Tax rates are caller-supplied data — never
-determined; rendering stays out (HTML is a `render` recipe, PDF
-consumer-side); no dunning, no e-invoicing formats, no subscription or
-pricing logic (the billing anti-scope stands).
-
-Deps: `core/money`; `core/fsm`.
-
 ## async/
 
 ---
--- a/finance/invoice/bench_test.go
+++ b/finance/invoice/bench_test.go
@@ -1,0 +1,113 @@
+package invoice_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/dmitrymomot/forge/core/decimal"
+	"github.com/dmitrymomot/forge/core/money"
+	"github.com/dmitrymomot/forge/finance/invoice"
+)
+
+func benchLines(n int) []invoice.LineItem {
+	rates := []decimal.Decimal{
+		decimal.MustParse("0"),
+		decimal.MustParse("0.07"),
+		decimal.MustParse("0.19"),
+	}
+	lines := make([]invoice.LineItem, n)
+	for i := range lines {
+		lines[i] = invoice.LineItem{
+			Description: fmt.Sprintf("line %d", i),
+			Quantity:    decimal.FromInt(int64(1 + i%3)),
+			UnitPrice:   money.FromMinor(int64(999+i*137), money.EUR),
+			TaxRate:     rates[i%len(rates)],
+		}
+	}
+	return lines
+}
+
+func BenchmarkCompute(b *testing.B) {
+	for _, n := range []int{1, 10, 100} {
+		lines := benchLines(n)
+		b.Run(fmt.Sprintf("per_line/%d", n), func(b *testing.B) {
+			b.ReportAllocs()
+			for b.Loop() {
+				if _, err := invoice.Compute(lines, invoice.RoundPerLine); err != nil {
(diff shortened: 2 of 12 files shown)


dmitrymomot:
Addressed all review findings in de19bfe (plus one self-review catch in 504ae9f):

- **Successive partial payments** (self-review): a second partial payment targeted the document's current status and the lifecycle has no self-edge, so it failed with an illegal transition. Only actual moves fire the machine now; regression test proven to fail pre-fix.
- **`Party.Address` aliasing** (both reviews): `NewCreditNote` now clones the address slices, with a test alongside the lines-independence one.
- **Sub-minor-unit payments** (agent review, Important): `ApplyPayment` rejects amounts not representable in the currency's minor units — 3 × 33.333 against a 100.00 total would strand the document in `partially_paid` forever while `Outstanding()` renders 0.00. Rejected rather than rounded so the recorded amount stays equal to the ledger posting the ref points at.
- **Misleading error paying a non-payable document**: status eligibility is now checked before the money math, so a draft reports `fsm.ErrIllegalTransition` instead of `ErrOverpayment`/currency errors. The replay scan stays ahead of it, so at-least-once feeds redelivering the final payment of a now-paid document still no-op.
- **Uniform void error shape**: added the guarded `partially_paid → void` edge, so every void-with-payments path reports `fsm.ErrGuardDenied` wrapping `ErrHasPayments`.
- **`Void` zero timestamp**: rejected with `ErrZeroTime` (generalized from `ErrIssueTime`).
- Tests added for each, plus the already-overdue re-sweep contract.

On the non-blocking notes: a transactional (postgres) `SequenceStore` driver is a natural follow-up once `finance/ledger` lands; the zero-total-invoice corner (issues fine, can never reach `paid`) is left as a conscious degenerate case — consumers void or ignore them.

PR #100, opened 2026-07-20
perf(decimal): int64 fast path for scale-reducing Rescale

dmitrymomot:
## Summary

`Decimal.Rescale`'s scale-reduction branch always promoted to `math/big` — `roundBig(d.bigCoef(), …)` — even when the coefficient was on the int64 fast path, so every currency rounding (`Round`, `Truncate`, `Floor`, `Ceil`) allocated. Profiling `finance/fxrate`'s `BenchmarkSnapshotConvertDirect` (PR #96) showed ~100% of its 11 allocs/op came from `Rescale → roundBig → big.nat.make/pow10Big/NewInt`.

This adds `roundInt64`, an allocation-free mirror of `roundBig`: divide by `10^drop` from a package `pow10Int64` table (`drop ≤ 18`), apply the mode to the remainder. `MinInt64` (magnitude overflows int64 negation) and `drop ≥ 19` defer to the retained big path, which is unchanged. The half-mode tie compare `2*r` vs `div` cannot overflow (`r < div ≤ 10^18`).

## Correctness

- White-box differential grid: 38 boundary coefficients (incl. `MaxInt64`, `MinInt64±1`, tie points ±5/±15/±25/±50/±150/±250) × drops 0–20 × all 7 `RoundingMode`s, asserting `roundInt64 == roundBig`.
- New fuzz target `FuzzRoundInt64VsBig` (same differential): 8.7M execs / 30s, clean.
- Full package tests green with `-race`; `core/money` (main rounding consumer) green.

## Benchmarks (Apple M3 Max, benchstat n=6, back-to-back same machine)

```
Round      190.95n ± 2%  →   3.23n ± 14%   −98.3%   328 B/op, 12 allocs → 0/0
Truncate   150.85n ± 1%  →   3.22n ±  2%   −97.9%   232 B/op, 10 allocs → 0/0
RoundBig   234.6n  ± 1%  →  219.3n ±  7%   (big-backed path: code unchanged)
RescaleUp  4.457n  ± 1%  →  4.341n ±  6%   (scale-increase path: unchanged)
```

New `BenchmarkRescaleDown` covers the scale-reduction path directly: 3.20n, 0 allocs.

Downstream: `finance/fxrate`'s direct-convert path drops from 11 allocs/op to ~0 once this lands (its profile was entirely `Rescale`).

### DIFF
--- a/core/decimal/decimal_bench_test.go
+++ b/core/decimal/decimal_bench_test.go
@@ -139,6 +139,14 @@ func BenchmarkRescaleUp(b *testing.B) {
 	}
 }
 
+func BenchmarkRescaleDown(b *testing.B) {
+	d := decimal.MustParse("1234.56789012")
+	b.ReportAllocs()
+	for b.Loop() {
+		_ = d.Rescale(2, decimal.HalfEven)
+	}
+}
+
 func BenchmarkRescaleBig(b *testing.B) {
 	d := decimal.MustParse(bigLit)
 	b.ReportAllocs()
--- a/core/decimal/round.go
+++ b/core/decimal/round.go
@@ -1,6 +1,9 @@
 package decimal
 
-import "math/big"
+import (
+	"math"
+	"math/big"
+)
 
 // RoundingMode selects how discarded low-order digits are handled when reducing
 // scale. HalfEven (banker's rounding) is the default and zero value.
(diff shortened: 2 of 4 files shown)


dmitrymomot:
Addressed both review notes: added keep-in-sync comments on both `roundAwayFromZero` variants (pointing at the differential tests as the enforcement), and codified the white-box-test exception in `docs/design.md` — black-box only, with `package foo` test files as the narrow exception for pinning unexported primitives' behavior (this file, the mmdb decoder precedent).

dmitrymomot:
**Overlap notice:** open PR #95 (finance/formula) carries a near-identical `Rescale` int64 fast path (same `roundInt64`/`pow10Int64`/`roundAwayFromZeroInt64` mirror-of-`roundBig` design, MinInt64 + drop≥19 fallback) plus a `Cmp` fast path and `RoundingMode` text marshaling its package needs — the implementations are semantically equivalent, convergent by construction.

Recommendation: **land #95 first**, then this PR rebases down to its unique residue — the white-box differential grid test asserting `roundInt64 == roundBig` across boundary coefficients × drops 0–20 × all 7 modes, `FuzzRoundInt64VsBig`, `BenchmarkRescaleDown`, the keep-in-sync comments, and the design.md white-box-exception note — or closes if that residue isn't wanted. Will handle whichever way on request.

PR #105, opened 2026-07-20
ops/tracing: Tracer seam, W3C trace-context middleware, otel driver

dmitrymomot:
Implements the `ops/tracing` catalog entry: a minimal distributed-tracing facade with a `Tracer`/`Span` seam, W3C trace-context middleware, a `trace_id` log extractor, outbound header propagation pairing with `httpclient`, and an isolated `tracing/otel` driver. The entry is removed from `docs/packages.md` per the shipped-package rule.

## Design

- **Seam** (`tracing.go`): `Tracer.Start` + `Span` interfaces using plain `slog.Attr` for attributes — same vocabulary as logging, no new KV type. `StartConfig`/`NewStartConfig` are exported so out-of-tree drivers can apply `StartOption`s. Kinds and statuses mirror the OTel vocabulary for one-to-one driver mapping.
- **W3C trace context** (`spancontext.go`): zero-alloc `ParseTraceparent` enforcing the spec (lowercase hex only, non-zero ids, version `ff` rejected, future versions tolerated), `SpanContext.Traceparent()` formatter, opaque `TraceState` pass-through.
- **Zero-dependency default** (`tracer.go`): `New()` is propagation-only — continues the inbound trace (or mints a new one) with a fresh span id per hop but records nothing, so `trace_id` log correlation and downstream propagation work with no exporter. New roots are marked sampled so parent-based samplers downstream aren't suppressed. `NewNoop()` returns ctx unchanged, which keeps an inbound remote parent visible to logs and propagation — a disabled service still forwards the trace it received.
- **Middleware** (`middleware.go`): parses inbound `traceparent`/`tracestate` into a remote parent (`WithTrustInbound(false)` for public edges), wraps the handler in a `KindServer` span named after the matched `r.Pattern` (`WithPathFunc` hook for other routers), records `http.request.method`, `url.path`, `http.route`, `http.response.status_code`, marks 5xx `StatusError`, and ends the span even on panic. `WithContextAttr` is the construction-time tenancy seam (fail-closed `"unknown"`, mirrors `metrics.WithContextLabel`); single-tenant apps skip it and pay nothing.
- **Log + outbound pairing** (`context.go`, `propagation.go`): `LogExtractor` adds `trace_id` from the live span or remote parent; `PropagationHeaders` is shaped for `httpclient.WithContextHeaders`, `Inject` covers other transports and queue hand-off.
- **otel driver** (`tracing/otel`): adapts any `trace.TracerProvider`; seeds a forge remote parent into the OTel context slot only when no live OTel span exists (third-party OTel instrumentation composes both ways), flattens slog groups to dotted keys, resolves `LogValuer`s, stringifies uint64 overflow. The OTel SDK is a test-only dependency; `go.opentelemetry.io/otel`+`/trace` move from indirect to direct.

## Benchmarks (Apple M-series, `-benchmem`)

```
BenchmarkParseTraceparent-14          64.98 ns/op     0 B/op   0 allocs/op
BenchmarkTraceparent-14               30.70 ns/op    64 B/op   1 allocs/op
BenchmarkStartChild-14                80.22 ns/op   176 B/op   4 allocs/op
BenchmarkSpanContextFromContext-14    13.15 ns/op     0 B/op   0 allocs/op
BenchmarkLogExtractor-14              31.08 ns/op    32 B/op   1 allocs/op
BenchmarkPropagationHeaders-14       139.3  ns/op   480 B/op   4 allocs/op
BenchmarkMiddleware-14               389.0  ns/op   912 B/op  14 allocs/op
otel: BenchmarkStartEnd-14           381.4  ns/op   704 B/op   8 allocs/op
otel: BenchmarkSetAttributes-14      169.6  ns/op   356 B/op   2 allocs/op
```

Post-benchmark pass: the hot parse path is already zero-alloc and the remaining allocations are irreducible boxing/context/string costs (span interface value, `context.WithValue`, hex string, header map), so no perf-motivated complexity was added — numbers above are the shipped baseline.

## Testing

Black-box tests throughout: spec-negative parser matrix (uppercase hex, zero ids, version `ff`, future-version tolerance), identity inheritance/remote adoption/new-root, noop pass-through, middleware (route naming, inbound trust on/off, oversized-tracestate drop, 5xx status, panic still ends span, tenancy attr fail-closed), end-to-end hop propagation, and otel driver tests against the SDK's in-memory exporter (remote parent seeding, parent-based sampling of unsampled inbound, attribute conversion incl. nested groups and uint64 overflow, middleware integration). `just lint` clean; full `go test -race ./...` green.

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -638,15 +638,6 @@ Deps: none (stdlib only).
 
 ---
 
-**ops/tracing**
-
-`Tracer` seam, W3C traceparent middleware, trace_id log extractor;
-`tracing/otel` driver isolated. Pairs with httpclient propagation.
-
-Deps: none forge-internal (otel external, isolated).
-
----
-
 **ops/secretsource**
 
 Hot-reloadable secrets behind a `Provider` seam, exposing `redact.Secret`
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,9 @@ require (
 	github.com/testcontainers/testcontainers-go/modules/postgres v0.43.0
 	github.com/testcontainers/testcontainers-go/modules/redis v0.43.0
 	go.mongodb.org/mongo-driver/v2 v2.7.0
+	go.opentelemetry.io/otel v1.44.0
+	go.opentelemetry.io/otel/sdk v1.44.0
+	go.opentelemetry.io/otel/trace v1.44.0
 	golang.org/x/crypto v0.53.0
 	golang.org/x/text v0.38.0
 	gopkg.in/yaml.v3 v3.0.1
@@ -309,9 +312,7 @@ require (
 	go.augendre.info/fatcontext v0.9.0 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.69.0 // indirect
-	go.opentelemetry.io/otel v1.44.0 // indirect
 	go.opentelemetry.io/otel/metric v1.44.0 // indirect
-	go.opentelemetry.io/otel/trace v1.44.0 // indirect
 	go.uber.org/atomic v1.11.0 // indirect
 	go.uber.org/automaxprocs v1.6.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
--- a/ops/tracing/bench_test.go
+++ b/ops/tracing/bench_test.go
@@ -1,0 +1,88 @@
+package tracing_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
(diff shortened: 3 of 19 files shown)


dmitrymomot:
Addressed both review findings in 8d6f391:

- **http.request.method_original**: non-standard verbs still normalize to `OTHER` in the span name and `http.request.method` for cardinality control, but the raw method now rides a `http.request.method_original` attribute (only added when normalization changed it), matching OTel semconv. Covered by `TestMiddlewareNonStandardMethodNormalized` + a no-attr-for-standard-methods test.
- **Inject nil-header**: `Inject` now no-ops on a nil header instead of panicking, and the doc comment states it, so "always safe to call" is accurate. Covered in `TestInject`.

Both threads resolved; tests and lint green.

PR #108, opened 2026-07-21
web/risk: generic anti-fraud gate with scorer strategies and probability threshold

dmitrymomot:
## What

New package `web/risk`: a generic anti-fraud gate. Consumer-supplied scorer functions each return a fraud probability in [0, 1] for an input `T`, a combining strategy folds them into one score, and a threshold gate decides whether the call proceeds. The package owns combining, gating, and error flow only — all fraud logic (useragent checks, fingerprint mismatch, velocity, geo) stays in consumer scorers wired from `web/useragent` / `web/fingerprint` / `web/geoip` / `web/clientip`. Design doc: `docs/superpowers/specs/2026-07-21-web-risk-design.md`.

## API

- `New[T](WithScorer(...), WithGate(0.8), ...)` → immutable, concurrent-safe `Engine[T]`; `Check(ctx, input) error` (nil = proceed), `Score(ctx, input)` (no gating — telemetry / shadow tuning).
- Strategies: `Max` (default — fraud signals are not additive, one strong signal must not be diluted), `Mean`, `WithWeights(1, 3)` normalized weighted average (arity validated in `New`, which is why it is an option, not a `Strategy` value), or any custom `func([]float64) float64`.
- Gate trip (`score >= gate`, boundary inclusive): with no handler, returns `*FraudError` carrying per-scorer attribution (`Peak`/`PeakIdx`/`Scores`), matchable via `errors.Is(err, ErrFraud)`. With `OnFraud`, the handler's return decides: nil proceeds (shadow mode / divert-and-allow), error blocks verbatim.
- `Middleware(engine, buildInput, ...)` adapts to `net/http` per the sibling middleware pattern: default plain 403 on any `Check` failure (fraud and scorer infrastructure errors alike — fail closed), `WithRejectHandler` seam for decoy redirects / problem responses, `WithLogger` defaulting to `logger.NewNope()`.

## Semantics worth reviewing

- **Fail closed everywhere:** scorer error, NaN, ±Inf, or out-of-range score → check fails with the underlying error, never clamped or skipped (a NaN comparing false against the gate is the `auth/lockout` NaN hole). Custom strategy output is validated the same way; built-ins clamp float rounding only.
- **All scorers always run, in registration order, no short-circuit** — attribution stays complete and consumer scorer side effects don't vary with registration order. `ctx.Err()` checked between scorers.
- **Tenancy** is a passed value, not a seam (`core/phone` precedent): pure compute, no storage — tenant travels inside `T` or the consumer builds per-tenant engines.
- Fits the seam `web/smartlink` left consumer-side: a `Decider` decorator calling `Check` (pure scorers) for fraud diversion, `Middleware` in front of the redirect handler for I/O scorers, `Score` in `OnHit` for shadow tagging.

## Benchmarks (Apple M3 Max, before → after pooled scratch buffer)

| Benchmark | baseline | after | allocs |
|---|---|---|---|
| CheckPass/1scorer | 14.27 ns/op | 14.10 ns/op | 1 → **0** |
| CheckPass/4scorers | 25.86 ns/op | 21.67 ns/op | 1 → **0** |
| CheckPass/16scorers | 64.25 ns/op | 59.49 ns/op | 1 → **0** |
| CheckPassMean | 25.97 ns/op | 21.82 ns/op | 1 → **0** |
| CheckPassWeighted | 25.20 ns/op | 21.35 ns/op | 1 → **0** |
| CheckTrip | 46.74 ns/op | 55.68 ns/op | 2 → 2 |
| Score | 25.81 ns/op | 25.92 ns/op | 1 → 1 |
| MiddlewarePass | 32.28 ns/op | 26.82 ns/op | 1 → **0** |

Optimization: `Check` evaluates into a `sync.Pool`-ed scratch slice; the pass path (every legitimate request) allocates nothing, and `Score.Scores` attribution is cloned out of the pooled buffer only on a trip. Trip path pays ~9 ns for the pool round-trip + clone — the rare path by construction.

## Tests

Black-box in `risk_test`: `New` validation matrix, gate boundary, handler nil-proceeds vs error-blocks-verbatim, no-short-circuit ordering, fail-closed matrix (scorer error / NaN / ±Inf / out-of-range / custom-strategy NaN / ctx cancellation), strategy math + weight normalization, middleware pass/403/override/shadow/panic paths, fuzz on the weighted-average invariant (3.3M execs clean), runnable example. `-race` clean, `just lint` clean (incl. nilaway).

### DIFF
--- a/docs/superpowers/specs/2026-07-21-web-risk-design.md
+++ b/docs/superpowers/specs/2026-07-21-web-risk-design.md
@@ -1,0 +1,132 @@
+# web/risk — Design
+
+**Date:** 2026-07-21
+**Status:** Approved
+
+## Purpose
+
+Generic anti-fraud gate: consumer-supplied scorer functions each return a fraud probability for an input, a pluggable strategy combines them, and a threshold gate decides whether the call proceeds, trips a fraud handler, or errors. Mechanism only — the package owns combining, gating, and error flow; all fraud *logic* (useragent checks, fingerprint mismatch, velocity, geo rules) lives in consumer scorers. Fills the consumer-side "fraud diversion / bot filtering" seam that `web/smartlink` explicitly leaves to callers, and is equally usable mid-handler (signup, deposit, bonus claim) or off-request (queue consumers).
+
+Not `auth/abac` (boolean predicate policy → allow/deny): `risk` is graded scoring with a threshold and an action hook. Not `web/ipfilter` (static list gate). If a consumer only needs boolean rules, they should use `abac`, not this.
+
+## Decisions (from brainstorming Q&A)
+
+1. **Multiple scorers + combining strategy**, not a single function. Default strategy `Max` — fraud signals are not additive; one 0.9 signal must not be diluted by innocent 0.1s in an average. `Mean` and `Weighted` ship as alternatives.
+2. **Generic transport-free core + thin `net/http` middleware adapter in the same package** (option b from Q&A). Core never imports `net/http`; the adapter follows the `web/tenant` / `auth/guard` middleware pattern.
+3. **Fraud handler decides the outcome.** On gate trip with an `OnFraud` handler set, the handler runs and its return value controls flow: `nil` → proceed (shadow mode / divert-and-allow), error → blocked with that error. No handler → blocked with `ErrFraud`.
+4. **Fail closed, no fail-open option in v1.** Scorer error, NaN, or out-of-range score fails the check. Consumers wanting lenient scorers wrap their own recover-to-zero.
+5. **Name `web/risk`** — short, mechanism-named like `fingerprint`/`ipfilter`/`lockout`.
+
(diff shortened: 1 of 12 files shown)


dmitrymomot:
### Review feedback addressed (19ed222)

- Fixed `WithLogger(nil)` clobbering the `logger.NewNope()` default — now nil-guarded per the repo-wide middleware option convention (`web/tenant`, `web/attribution`, `web/fingerprint`, `web/smartlink`, `web/recoverer`).
- Added the suggested logger coverage: tests now pin the Warn record with `score`/`peak_scorer`/`path` attribution on the fraud path, the Error record on the scorer-infrastructure path, and the `WithLogger(nil)` no-panic case.

`go test -race`, `just lint` (incl. nilaway), and modernize all clean.

@claude ready for re-review.

PR #109, opened 2026-07-21
realtime/fanout: in-process pub/sub hub with pgbus/redisbus backplanes

dmitrymomot:
## Summary

Implements the `realtime/fanout` catalog entry: the ephemeral, at-most-once fan-out hub behind the upcoming `realtime/sse` and `realtime/presence` packages, plus both backplane drivers.

### Core (`realtime/fanout`)

- **Hub** — `New(...Option)`, `Publish(ctx, topic, payload)`, `Subscribe(ctx, topics, ...SubscribeOption)` returning a channel-backed `Subscription`. Publishers never block; publishing to a topic with no subscribers is a no-op. Topic registry entries are garbage-collected when their last subscriber leaves.
- **Explicit slow-consumer policy** — per-subscription bounded buffer with `DropOldest` (default), `DropNewest`, or `CloseSlow` (tears the subscription down; `Err()` reports `ErrSlowConsumer`). Every loss is counted on `Subscription.Dropped()`.
- **Replay ring** — `WithReplay(n)` keeps a per-topic ring of the newest n messages; `WithResumeAfter(id)` replays the gap and goes live seamlessly (the Last-Event-ID currency for `realtime/sse`). Rings of subscriber-less topics are retained for `WithReplayTTL` (default 5m) and swept amortized from the publish path — no background goroutine anywhere in the core.
- **Bus seam** — with `WithBus`, publishes route through the backplane only and local delivery happens on the bus receive path, so every instance observes one ordering and one loss profile (no self-echo dedup needed).
- **Tenancy** — fail-closed `WithScope` hook namespaces topics via a reserved separator (`0x1F`, rejected in topics and scopes), isolating tenants including across a shared bus. Single-tenant apps configure nothing.

### Drivers

- **`fanout/pgbus`** — LISTEN/NOTIFY over the caller's `*pgxpool.Pool`. JSON+base64 envelope on one configurable channel; the 8000-byte NOTIFY cap is enforced (`ErrPayloadTooLarge`). `supervisor.Service` receive loop with exponential-backoff reconnect; the LISTEN connection is closed on exit, never released back into the pool with LISTEN state attached.
- **`fanout/redisbus`** — Redis Pub/Sub on the caller's `goredis.UniversalClient`. NUL-framed binary envelope (zero encoding overhead). `ReceiveMessage` ignores ctx cancellation, so shutdown closes the PubSub via `context.AfterFunc` — found by the integration tests hanging on teardown.

### Testing

- Black-box unit suite: policies, replay/resume (incl. buffer-overflow preload and TTL sweep via mock clock), scope isolation, fake-bus wiring, hub/subscription lifecycle, `-race` stress.
- `TestCloseSlowDuringSubscribe` is a proven regression test: a publisher-triggered `CloseSlow` teardown in the window before `Subscribe` returned raced on `sub.states`; the test detects the race pre-fix under `-race` and passes post-fix (assignment moved before registration).
- Integration tier (`-tags=integration`, testcontainers): cross-instance publish/receive incl. self-delivery and binary payload round-trip, channel isolation, oversized-payload rejection (pgbus), and hub-to-hub end-to-end over each driver. Channels are uniquified per test.

### Benchmarks (M3 Max, required pass)

```
BenchmarkPublish/1subs-14           9209701     118.3 ns/op       0 B/op    0 allocs/op
BenchmarkPublish/8subs-14           1823379     652.2 ns/op       0 B/op    0 allocs/op
BenchmarkPublish/64subs-14           125995      8915 ns/op       0 B/op    0 allocs/op
BenchmarkPublishNoSubscribers-14  125275588     9.602 ns/op       0 B/op    0 allocs/op
BenchmarkPublishReplay-14           5171304     231.7 ns/op       0 B/op    0 allocs/op
BenchmarkPublishScoped-14           6036384     198.6 ns/op      16 B/op    1 allocs/op
BenchmarkSubscribeClose-14          2352213     509.4 ns/op    3752 B/op    9 allocs/op
BenchmarkResume-14                   544324      2165 ns/op   11000 B/op   13 allocs/op
```

Post-benchmark optimization pass: the hub-closed check on `Publish` moved from an RLock pair to an `atomic.Bool` (written under the hub mutex), 120.5→115.6 ns/op on the 1-sub path; the publish hot path is zero-alloc (scoped publish pays one 16 B key concat, documented).

### Catalog

`realtime/fanout` entry deleted per the ship rule; the `(planned)` markers on `realtime/sse` and `realtime/presence` deps cleared.

### DIFF
--- a/docs/packages.md
+++ b/docs/packages.md
@@ -192,19 +192,7 @@ handling, Last-Event-ID resume via the replay ring. The low-level writer
 stays exported as the brick under `web/htmx`'s SendComponent bridge and
 `llm.SSE`. Requires httpserver WriteTimeout=0 (documented).
 
-Deps: `realtime/fanout` (planned).
-
----
-
-**realtime/fanout**
-
-In-process pub/sub hub (bounded buffers, explicit slow-consumer policy)
-with the `Bus` seam for multi-instance backplanes and an optional bounded
-`WithReplay(n)` ring. Drivers: `fanout/pgbus` (LISTEN/NOTIFY — multi-
-instance push with zero new infrastructure), `fanout/redisbus` (takes the
-caller's `data/redis` client).
-
-Deps: drivers ride `data/postgres`, `data/redis`.
+Deps: `realtime/fanout`.
 
 ---
 
@@ -225,7 +213,7 @@ Deps: `ops/supervisor` (coder/websocket external).
 Who-is-here tracking with TTL heartbeats over `fanout` + `cache` — online
 indicators, concurrent-viewer counts.
 
-Deps: `resilience/cache`; `realtime/fanout` (planned).
+Deps: `resilience/cache`; `realtime/fanout`.
 
 ## auth/
 
--- a/realtime/fanout/bench_test.go
+++ b/realtime/fanout/bench_test.go
@@ -1,0 +1,148 @@
+package fanout_test
+
+import (
+	"context"
+	"strconv"
+	"testing"
+
+	"github.com/dmitrymomot/forge/realtime/fanout"
+)
+
+// drain consumes a subscription as fast as possible until its channel
+// closes.
+func drain(sub *fanout.Subscription) {
+	for range sub.C() {
+	}
+}
+
+func BenchmarkPublish(b *testing.B) {
+	ctx := context.Background()
+	payload := []byte(`{"text":"hi"}`)
+
+	for _, subs := range []int{1, 8, 64} {
+		b.Run(strconv.Itoa(subs)+"subs", func(b *testing.B) {
+			hub, err := fanout.New(fanout.WithDefaultBuffer(1024))
+			if err != nil {
+				b.Fatal(err)
+			}
+			defer hub.Close()
+			for range subs {
+				sub, err := hub.Subscribe(ctx, []string{"t"})
+				if err != nil {
+					b.Fatal(err)
+				}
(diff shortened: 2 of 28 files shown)


dmitrymomot:
Addressed the review: `ErrPayloadTooLarge` is now covered in the unit tier (`TestPublishPayloadTooLarge` using the lazy `testPool` helper — the size check runs before any pool I/O), and the redundant integration-tier duplicate was removed. Thread resolved.

dmitrymomot:
### Review feedback addressed (round 2 — minor observations)

- Fixed `Hub.Publish` error precedence: the closed check now runs before topic validation and scope resolution, so a closed hub reports `ErrClosed` even when the scope hook would also error. Pinned by a new `closed hub wins over scope errors` test.
- Added the `pgbus` doc callout that `Run` reserves one pool connection for LISTEN — size `pgxpool.Pool.MaxConns` accordingly on pools shared with application traffic.

Both non-blocking observations from the original review are now addressed. @claude ready for re-review.

dmitrymomot:
Fixed the `Subscribe` asymmetry in `c1ff6cc`: the closed check now runs first there too (the under-lock check stays authoritative), so `ErrClosed` wins over scope errors on both `Publish` and `Subscribe`. Pinned in the same `closed hub wins over scope errors` test. Thread resolved.

PR #113, opened 2026-07-21
perf(pgstore): sweep the optional-filter OR idiom that defeats indexes under generic plans

dmitrymomot:
Repo-wide sweep of the `($n = '' OR col = $n)` / `($n IS NULL OR ...)` optional-filter SQL idiom, following its removal from ops/approval/pgstore in #112 (commit 0f57e7c).

## Why

pgx prepares every query, and Postgres switches a prepared statement to a **generic plan after five executions**. A generic plan cannot prune the ORs, so every query built with the idiom degrades to a scan that ignores the very indexes the migrations create for it. This bites exactly in production (warmed statements, real table sizes) while staying invisible in tests.

## What changed

| package | query | old (generic plan) | new (generic plan) |
|---|---|---|---|
| auth/apikey/pgstore | List (tenant+subject) | Parallel Seq Scan, 2634 buffers, 10.2ms | Bitmap Index Scan on `forge_api_keys_list_idx`, 406 buffers, 0.29ms |
| auth/oauthserver/pgstore | List (tenant) | Seq Scan, 190k rows filtered, 25.9ms | Bitmap Index Scan on `oauth_clients_tenant_idx`, 17.3ms¹ |
| web/smartlink/pgstore | List (tenant, LIMIT 100) | Parallel Seq Scan, 1517 buffers, 7.3ms | Index Only Scan, 36 buffers, 0.05ms |
| ops/auditlog/pgsink | List (tenant, LIMIT 51) | pkey walk + filter, 1152 buffers, 0.70ms | Index Only Scan on `forge_audit_events_list_idx`, 54 buffers, 0.03ms |
| ops/auditlog/pgsink | Verify batch @ offset 5000 | 5545 buffers, 5001 rows removed by filter — **O(N²/batch) over the stream** | 507 buffers flat, 0.09ms |

¹ both sides return the tenant's full 10k rows; the win grows with table size and shrinks with result size.

All numbers: EXPLAIN (ANALYZE) on postgres:18, 200k rows, `SET plan_cache_mode = force_generic_plan` (the state warmed pgx statements reach), PREPARE/EXECUTE with the exact old vs new query shapes.

## How

- **apikey, pgsink List**: WHERE built from only the filters actually set (the #112 pattern — placeholder positions string-built, all values bound). Statement-cache bound: 4 and 2⁸ shapes respectively.
- **oauthserver, pgsink Verify**: only one optional predicate, so two static statements instead of dynamic SQL. Verify's keyset bound is the important one: as a filter it rescans the chain from genesis every batch; as an index condition each batch starts at the bound.
- **smartlink**: List gets the conditional predicate; the Deactivate/Activate/Delete point statements become unscoped/tenant-scoped pairs. Perf is not the driver for those three — `code` is the primary key, so they are point lookups under any plan — they're included so the package is free of the idiom and it can't be copied back into a query where it does matter.

Not touched: `ops/approval/pgstore` (fixed in #112), and column-level NULL checks like `last_used_at IS NULL OR last_used_at < $3` (auth/totp, async/queue, resilience/ratelimit) — those are genuine SQL semantics on column values, not parameter-driven optional filters, and a generic plan handles them fine.

## Verification

- `go test -race` on all four package trees: pass
- `go test -tags integration -race` (testcontainers, live Postgres) on all four pgstores: pass — behavior pinned identical by each package's existing conformance suites
- `just lint` (golangci-lint + nilaway, integration tags): 0 issues

### DIFF
--- a/auth/apikey/pgstore/pgstore.go
+++ b/auth/apikey/pgstore/pgstore.go
@@ -86,14 +86,13 @@ func (s *Store) GetByHash(ctx context.Context, hash string) (apikey.Key, error) 
 
 // List returns records matching f, newest first (UUIDv7 ids are
 // time-ordered, so id DESC is creation order).
-//
-// The WHERE clause carries only the filters actually set, rather than the
-// static `($n = ” OR col = $n)` idiom: once a prepared statement switches
-// to a generic plan (pgx prepares every query, and Postgres goes generic
-// after five executions) those ORs cannot be pruned and the planner stops
-// using forge_api_keys_list_idx. The filter combinations bound the
-// statement cache at 4 shapes.
 func (s *Store) List(ctx context.Context, f apikey.Filter) ([]apikey.Key, error) {
+	// The WHERE clause carries only the filters actually set, rather than
+	// the static `($n = '' OR col = $n)` idiom: once a prepared statement
+	// switches to a generic plan (pgx prepares every query, and Postgres
+	// goes generic after five executions) those ORs cannot be pruned and
+	// the planner stops using forge_api_keys_list_idx. The filter
+	// combinations bound the statement cache at 4 shapes.
 	var (
 		conds []string
 		args  []any
--- a/auth/apikey/pgstore/pgstore.go
+++ b/auth/apikey/pgstore/pgstore.go
@@ -5,6 +5,8 @@ import (
 	"embed"
 	"errors"
 	"io/fs"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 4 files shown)


dmitrymomot:
Fixed both inline findings in adeabd2:

- **Curly-quote typo**: root cause was gofmt itself — since Go 1.19 it rewrites doc comments via go/doc/comment, which applies the old godoc quoting convention and converts a literal `''` to `”`. Moved the idiom explanations into function bodies (where gofmt leaves comments untouched, matching pgsink's existing style), so the SQL stays verbatim and the formatter can't re-mangle it.
- **Statement-cache pressure**: expanded the pgsink List comment with the math — 2^8 = 256 worst-case shapes is half of pgx's default 512-entry per-connection LRU, reached only if one connection sees every combination (a paged UI uses a handful), and an eviction costs one re-prepare round-trip, not an error.

On the non-blocking notes: the PR body's #112 reference now says it's pending merge (the sweep covers main's state once both land), and the missing `bench_test.go` debt in these four packages predates this PR — worth its own pass.

PR #116, opened 2026-07-22
feat(auth/invite): single-use tenant invitations with token rotation

dmitrymomot:
## Summary

Implements the `auth/invite` catalog entry: email-addressed, single-use, expiring token invitations into a tenant, carrying an opaque role payload the package never interprets. The catalog entry is removed from `docs/packages.md` per the shipped-package rule.

- **Token discipline (apikey mold):** `inv_`-prefixed secrets — 43-char base62 payload (~256 bits) + CRC32 checksum. Malformed input is rejected allocation-free before any store access; only the hex SHA-256 is persisted; the plaintext is returned exactly once (at Create, and at Resend which rotates it).
- **Constant-time, atomically single-use Accept:** hash lookup + `consttime.StringEqual` defense-in-depth, then a conditional store update (`WHERE accepted_at IS NULL AND revoked_at IS NULL AND expires_at > $2`) so exactly one concurrent Accept wins and replays report `ErrAlreadyAccepted`. Accept returns the verified `Claim{Tenant, Email, Role, ID}` — membership creation, seat limits, and the email send stay consumer-side.
- **Peek** verifies without consuming, so email scanners that prefetch links cannot burn an invite and the "join org" page renders before the user commits.
- **Revoke** is idempotent (keeps the original timestamp); revoking an accepted invite reports `ErrAlreadyAccepted`. **Resend** rotates hash + expiry — the old token is dead the moment resend succeeds, and an expired (but not accepted/revoked) invite can be revived.
- **Tenancy:** fail-closed `WithScope` seam confines Create/Get/List/Revoke/Resend (hook error or empty tenant → `ErrScope`; cross-tenant reads → `ErrNotFound`); Accept/Peek are deliberately unscoped because the invitee has no tenant context yet — the record carries the tenant. Single-tenant use pays zero ceremony.
- **Stores:** in-memory for tests/dev + `pgstore` pgx driver with embedded goose migration (`forge_invites`, unique hash index, dynamic WHERE in List — no `($n='' OR …)` idiom). The single-use/rotate guards are conditional single-row UPDATEs classified on zero rows affected, so no transactions are needed.

## Benchmarks (M3 Max, memory store)

```
BenchmarkCreate-14                	  862774	  1367 ns/op	  806 B/op	  9 allocs/op
BenchmarkPeek-14                  	 1000000	  1138 ns/op	 1376 B/op	 18 allocs/op
BenchmarkPeekMalformedReject-14   	17445250	 68.36 ns/op	    0 B/op	  0 allocs/op
BenchmarkListPending-14           	   80653	 14940 ns/op	18432 B/op	  1 allocs/op
```

Post-benchmark pass: the DoS-relevant reject path is 0 allocs (same crc32String trick as apikey). Peek/Accept's 18 allocs are entirely `consttime.StringEqual`'s deliberate double-HMAC pattern (verified via alloc profile) — identical to apikey's verify path and not worth bypassing for a human-rare operation, so no perf-motivated changes were made.

## Testing

- Unit + `-race`: manager lifecycle, rejection classes, concurrent single-use (exactly one winner of 32), scope fail-closed sweep across all five management ops, store contract (CAS classification, rotate reindexing, idempotent revoke), store-error wrapping, malformed-token-never-touches-store.
- `FuzzPeek` (15s clean): no forged token verifies, no input panics.
- Integration (`-tags=integration`, testcontainers Postgres): full store contract against real Postgres incl. concurrent accept and an end-to-end manager flow — all green locally.

### DIFF
--- a/auth/invite/bench_test.go
+++ b/auth/invite/bench_test.go
@@ -1,0 +1,73 @@
+package invite_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/dmitrymomot/forge/auth/invite"
+)
+
+func BenchmarkCreate(b *testing.B) {
+	mgr := invite.New(invite.NewMemoryStore())
+	ctx := context.Background()
+	b.ReportAllocs()
+	for b.Loop() {
+		if _, _, err := mgr.Create(ctx, invite.CreateParams{Email: "a@b.com", Tenant: "t1", Role: "editor"}); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+func BenchmarkPeek(b *testing.B) {
+	// The token-verification path minus the consuming write.
+	mgr := invite.New(invite.NewMemoryStore())
+	ctx := context.Background()
+	_, plaintext, err := mgr.Create(ctx, invite.CreateParams{Email: "a@b.com", Tenant: "t1"})
+	if err != nil {
+		b.Fatal(err)
+	}
+	b.ReportAllocs()
+	for b.Loop() {
+		if _, err := mgr.Peek(ctx, plaintext); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+func BenchmarkPeekMalformedReject(b *testing.B) {
+	// The DoS-relevant path: checksum rejection without store access.
+	// Target: zero allocations.
+	mgr := invite.New(invite.NewMemoryStore())
+	ctx := context.Background()
+	_, plaintext, err := mgr.Create(ctx, invite.CreateParams{Email: "a@b.com"})
+	if err != nil {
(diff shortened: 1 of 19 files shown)


dmitrymomot:
Addressed the review's List-index observation in-PR rather than deferring: added partial index `forge_invites_pending_idx (expires_at) WHERE accepted_at IS NULL AND revoked_at IS NULL` to the (unreleased) 00001 migration. Its predicate matches `List`'s pending conditions exactly, so the unscoped cross-tenant pending listing stays indexed as expired-but-pending rows accumulate — no query changes needed. Integration test extended to cover the unscoped pending-only shape. `just lint` + integration tests green locally.

PR #120, opened 2026-07-23
fix(realtime): SSE resume cursor, invalid events, send-timeout, retry overflow + fanout cross-topic ordering

dmitrymomot:
## Description

Fixes five issues from a review of `realtime/sse` (one dependency fix lands in `realtime/fanout`). A sixth reported issue — "idle stream fails after the first send timeout" — was verified as a **false positive** (the write deadline is re-armed before every write; empirically an idle-then-send succeeds) and is not addressed.

### `realtime/fanout` — cross-topic delivery ordering (High)
Hub IDs are a single global sequence but topics lock independently, so a subscriber multiplexing several topics could observe a higher ID enqueued ahead of a lower one and regress its `Last-Event-ID` cursor (skip on disconnect, or duplicate on reconnect). A hub-level `seqMu` now serializes ID allocation, ring insertion, and delivery. Single-topic publish is unaffected; concurrent multi-topic publish serializes what was parallel.

### `realtime/sse`
- **Stale cross-instance cursor (High).** The default resume cursor is now namespaced with a random per-instance epoch (`<epoch>.<seq>`). Hub IDs are per-instance and reset on restart, so a numeric cursor from another instance or a previous boot was mis-applied against this instance's ID space and silently dropped the messages it overlapped. A foreign epoch now degrades to a live-only stream, as the package docs already promised. Custom encoders keep raw-cursor resume.
- **Invalid event → reconnect loop (Medium).** `ErrInvalidEvent` from `Send` (e.g. a fanout topic containing CR/LF becomes an invalid event name) is logged-and-skipped instead of ending the stream; validation writes nothing, so the stream is intact. Previously, with replay enabled, the bad event was re-served on every reconnect and trapped clients in a loop.
- **Send-timeout silently unenforced (Medium).** When deadline control is unavailable (a flushable writer without `Unwrap`) and `sendTimeout > 0`, `NewWriter` now fails closed before committing rather than silently dropping stall protection. `WithSendTimeout(0)` opts into unbounded writes.
- **Retry rounding overflow (Low).** Divide-then-round-up so a duration near `MaxInt64` no longer overflows into a negative `retry:` field.

## Benchmarks (`realtime/fanout`, `-count=5`, 14 cores)

| bench | before | after | Δ |
|---|---|---|---|
| `Publish/1subs` | 122.4 ns | 121.3 ns | none |
| `Publish/8subs` | 684.8 ns | 685.6 ns | none |
| `Publish/64subs` | ~10.2 µs | ~9.6 µs | none (noise) |
| `PublishMultiTopic` (concurrent) | 260.7 ns | 413.6 ns | +153 ns (~1.6×) |

All 0 allocs. Single-topic publish (the common SSE shape) is unaffected — `seqMu` is uncontended there. Only concurrent multi-topic publishing pays, which is the cost of the monotonic cross-topic cursor.

## Reviewer notes
- The `seqMu` regression test (`TestMultiTopicDeliveryMonotonic`) was **proven to fail pre-fix** under `-race` (a higher ID delivered before a lower one); it passes with the fix.
- The fail-closed send-timeout is a behavior change: a plain `httptest.ResponseRecorder` (flushable, no deadline support) now requires `WithSendTimeout(0)`. Recorder-based tests and both examples were updated accordingly. Real server connections support deadlines, so production code keeps the default.
- Default resume cursors change format to `<epoch>.<seq>`; opaque to clients. In-flight clients holding old bare-numeric cursors degrade to live-only once on deploy, then resume normally.

## Checklist

- [x] `just lint` passes
- [x] `just test` passes
- [x] New code has tests

### DIFF
--- a/realtime/fanout/bench_test.go
+++ b/realtime/fanout/bench_test.go
@@ -44,6 +44,38 @@ func BenchmarkPublish(b *testing.B) {
 	}
 }
 
+// BenchmarkPublishMultiTopic drives concurrent publishes across several topics
+// — the workload seqMu serializes to keep cross-topic delivery ordered.
+func BenchmarkPublishMultiTopic(b *testing.B) {
+	ctx := context.Background()
+	payload := []byte(`{"text":"hi"}`)
+	topics := []string{"a", "b", "c", "d"}
+
+	hub, err := fanout.New(fanout.WithDefaultBuffer(1024))
+	if err != nil {
+		b.Fatal(err)
+	}
+	defer hub.Close()
+	for _, topic := range topics {
+		sub, err := hub.Subscribe(ctx, []string{topic})
+		if err != nil {
+			b.Fatal(err)
+		}
+		go drain(sub)
+	}
+
+	b.ReportAllocs()
+	b.RunParallel(func(pb *testing.PB) {
+		i := 0
+		for pb.Next() {
+			if err := hub.Publish(ctx, topics[i%len(topics)], payload); err != nil {
+				b.Fatal(err)
+			}
+			i++
+		}
+	})
+}
+
 func BenchmarkPublishNoSubscribers(b *testing.B) {
 	ctx := context.Background()
 	hub, err := fanout.New()
--- a/realtime/fanout/fanout.go
+++ b/realtime/fanout/fanout.go
@@ -41,6 +41,9 @@ type Hub struct {
 	nextID    atomic.Uint64
 	lastSweep atomic.Int64
 	mu        sync.RWMutex
+	// seqMu orders ID allocation and delivery across topics so a subscriber on
+	// several topics never observes IDs out of order. See dispatch.
+	seqMu sync.Mutex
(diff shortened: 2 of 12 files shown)


dmitrymomot:
### Review feedback addressed (045e394)

Both non-blocking items from the review are handled:

- **Stale comment (`writer.go` `Send`)** — reworded the per-`Send` `SetWriteDeadline` comment. It no longer says "best-effort falls back to the server timeout"; it now states that `NewWriter` fails closed when `sendTimeout > 0` and deadline control is unavailable, so the call is guaranteed supported, and the error is ignored because a genuine failure (e.g. connection already closed) surfaces on the next `Write`/`Flush`.
- **Coverage gap (correct epoch, malformed sequence)** — added `TestHandlerResumeMalformedSequence`, which learns the live epoch then reconnects with `<epoch>.notanum` and asserts it degrades to live-only (the `ParseUint`-failure branch).

`just lint` and `just test` pass locally.

@claude ready for re-review.

